PR #2, opened 2019-06-05
Adds ReLU ANN algorithm and basic user interface

ajaysub110:


tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,119 @@
...
+        if_init = {
+            "Vmem":genn_model.init_var("Uniform", {"min": self.Vres, "max": self.Vthr}),
+            "SpikeNumber":0
+        }
+        
+        cs_init = {"magnitude":10.0}
+
+        # Define model and populations
+        self.g_model = genn_model.GeNNModel("float","g_model")
+        self.neuron_pops = {}
```
neworderofjamie: I don't see much point in making these dictionaries, if they were lists you could just reference e.g. the last layer with ``self.neuron_pops[-1]``

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,119 @@
...
+                    "if"+str(i-1),tf_weights[i-1].shape[0],if_model,if_params,if_init
+                )
+
+            # Postsynaptic neuron
+            self.neuron_pops["if"+str(i)] = self.g_model.add_neuron_population(
+                "if"+str(i),tf_weights[i-1].shape[1],if_model,if_params,if_init
+            )
+
+            # Synapse
+            self.syn_pops["syn"+str(i-1)+str(i)] = self.g_model.add_synapse_population(
```
neworderofjamie: I'm not sure you need to store the synapse populations at all

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,119 @@
...
+class ReLUANN():
+    def __init__(self,neuron_resting_voltage=-60.0,neuron_threshold_voltage=-55.0,
+        membrane_capacitance=1.0, model_timestep=1.0,):
+        self.Vres = neuron_resting_voltage
+        self.Vthr = neuron_threshold_voltage
+        self.Cm = membrane_capacitance
+        self.timestep = model_timestep
+
+    def convert(self, tf_model):
+        # create custom classes
```
neworderofjamie: This function probably needs to check that tf_model is a keras ``Sequential`` as, otherwise I don't think it works. Otherwise throw a ``NotImplementedError``

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,119 @@
...
+        
+        self.current_source = self.g_model.add_current_source("cs",cs_model,"if0",{},cs_init)
+
+        self.g_model.dT = self.timestep
+        self.g_model.build()
+        self.g_model.load()
+
+        return self.g_model
+
+    def evaluate(self, X, y=None, single_example_time=350.):
```
neworderofjamie: It would probably be good if ``single_example_time`` was passed to the constructor - that way this function can be entirely generic

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,119 @@
...
+        y = y.reshape(n_examples)
+
+        runtime = n_examples * single_example_time
+        i = -1
+        n_correct = 0
+
+        while self.g_model.t < runtime:
+            if self.g_model.t >= single_example_time*(i+1):
+                # After example i -1,0,1,2,..
+                self.g_model.pull_var_from_device("if2",'SpikeNumber')
```
neworderofjamie: this would definitely be a place where being able to go ``self.neuron_pops[-1]`` would be better

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,119 @@
...
+        
+        cs_init = {"magnitude":10.0}
+
+        # Define model and populations
+        self.g_model = genn_model.GeNNModel("float","g_model")
+        self.neuron_pops = {}
+        self.syn_pops = {}
+
+        for i in range(1,n_layers): # 1,2 - synapses
+            if i == 1:
```
neworderofjamie: It would be good to add checks that the layers are compatible with what this algorithm supports i.e. no bias, dense weights, relu. If not, again, throw ``NotImplementedError``

tensor_genn/algorithms/relu_ann.py
```diff
@@ -87,29 +96,30 @@ def convert(self, tf_model):
...
+            if self.g_model.t >= self.single_example_time*(i+1):
                 # After example i -1,0,1,2,..
-                self.g_model.pull_var_from_device("if2",'SpikeNumber')
-                SpikeNumber_view = self.neuron_pops["if2"].vars["SpikeNumber"].view
+                self.g_model.pull_var_from_device("if"+str(n-1),'SpikeNumber')
+                SpikeNumber_view = self.neuron_pops[-1].vars["SpikeNumber"].view
                 n_correct += (np.argmax(SpikeNumber_view)==y[i])
                 i += 1
                 # Before example i 0,1,2,3,..
                 for j in range(len(self.neuron_pops)):
```
neworderofjamie: Here you could do ``for j, n in enumerate(self.neuron_pops):`` and save some typing

tensor_genn/algorithms/relu_ann.py
```diff
@@ -87,29 +96,30 @@ def convert(self, tf_model):
 
         return self.g_model
 
-    def evaluate(self, X, y=None, single_example_time=350.):
+    def evaluate(self, X, y=None):
```
neworderofjamie: Why is ``X`` upper-case and ``y`` lower case?

tensor_genn/algorithms/relu_ann.py
```diff
@@ -6,21 +6,31 @@
...
 
     def convert(self, tf_model):
+        # Check model compatibility
+        if not isinstance(tf_model,tf.keras.models.Sequential):
+            raise NotImplementedError('Implementation for type {} models not found'.format(type(tf_model)))
+        
+        for layer in tf_model.layers:
+            if not isinstance(layer,(tf.keras.layers.Dense,tf.keras.layers.Flatten)):
+                raise NotImplementedError('Only Dense layers are supported')
+        if len(tf_model.get_weights()) >= len(tf_model.layers):
```
neworderofjamie: How does this check work?

neworderofjamie:
This looks really good @ajaysub110! I'll just see if @tnowotny has any thoughts and then we can merge

tensor_genn/algorithms/relu_ann.py
ajaysub110: Hi Jamie,
I've implemented the checks for bias and dense layers. But the tf.keras API doesn't seem to have an activation function attribute that I can access. Any ideas on a workaround solution?
Thanks
Ajay

tensor_genn/algorithms/relu_ann.py
ajaysub110: That's the general convention that I've noticed for data (X) and labels (y). Should I change them to the same case?

tensor_genn/algorithms/relu_ann.py
ajaysub110: The Dense layer check just loops through each layer and checks the type.
Again (like activation fn), there is no direct way to access the use_bias parameter. So this checks if the number of weight matrices is greater than the number of synapses. If so, it means that atleast one synapse has more than 1 weight matrix i.e a bias exists too. 
Does this make sense?

neworderofjamie: I just had a quick look and you can access the ``activation`` and ``use_bias`` properties on ``Dense`` layers but not ``Flatten`` layers as they don't have them - I think it would be nicer to check the type of each layer in the loop as you encounter them and use these properties

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,129 @@
...
+        n = len(self.neuron_pops)
+        i = -1
+        n_correct = 0
+
+        while self.g_model.t < runtime:
+            if self.g_model.t >= self.single_example_time*(i+1):
+                # After example i -1,0,1,2,..
+                self.g_model.pull_var_from_device("if"+str(n-1),'SpikeNumber')
+                SpikeNumber_view = self.neuron_pops[-1].vars["SpikeNumber"].view
+                n_correct += (np.argmax(SpikeNumber_view)==y[i])
```
tnowotny: I don't understand this - you are adding to n_correct also when i == -1 (when no input was set and no simulation was run) by comparing to the last entry in y?
I think this needs a little bit of cleaning up.

tensor_genn/algorithms/relu_ann.py
tnowotny: I think this convention with capital X an small y is quite commonly used in ML. I would recommend to keep it.

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,129 @@
...
+        n_examples = len(X)
+        X = X.reshape(n_examples,-1)
+        y = y.reshape(n_examples)
+
+        runtime = n_examples * self.single_example_time
+        n = len(self.neuron_pops)
+        i = -1
+        n_correct = 0
+
+        while self.g_model.t < runtime:
```
tnowotny: I think between this condition and the >= condition in the next line, combined with the formula in line 104, you may be missing out on evaluating the outcome for the last example.

tensor_genn/algorithms/relu_ann.py
```diff
@@ -0,0 +1,129 @@
...
+
+        self.g_model.dT = self.timestep
+        self.g_model.build()
+        self.g_model.load()
+
+        return self.g_model
+
+    def evaluate(self, X, y=None):
+        n_examples = len(X)
+        X = X.reshape(n_examples,-1)
```
tnowotny: Sorry for being daft, but what is this doing?

ajaysub110: It converts the input X to the desired shape i.e. (num_examples, num_features). I've done this so that the method doesn't fail when images are passed as, say (60000, 28,28)

tensor_genn/algorithms/relu_ann.py
neworderofjamie: I did not know that! In that case, keep it

tensor_genn/algorithms/relu_ann.py
```diff
@@ -18,11 +18,15 @@ def convert(self, tf_model):
...
-                raise NotImplementedError('Only Dense layers are supported')
-        if len(tf_model.get_weights()) >= len(tf_model.layers):
-            raise NotImplementedError('Tensorflow model should be trained without biases')
+                raise NotImplementedError('Only Dense and Flatten layers are supported')
+            elif isinstance(layer, tf.keras.layers.Dense):
+                if layer.activation != tf.keras.activations.relu:
+                    print(layer.activation)
+                    raise NotImplementedError('Only ReLU activation function is supported')
+                if layer.use_bias == True:
+                    raise NotImplementedError('TensorFlow model should be trained without bias tensors')
```
neworderofjamie: :+1: much better

tensor_genn/algorithms/relu_ann.py
ajaysub110: I've implemented the checks for all layers except the last one since it may be used for classification and hence may use softmax/sigmoid

tensor_genn/algorithms/relu_ann.py
```diff
@@ -103,26 +107,27 @@ def evaluate(self, X, y=None):
...
+                    npop.vars["Vmem"].view[:] = random.uniform(self.Vres,self.Vthr)
                     self.g_model.push_state_to_device("if"+str(j))
                 
-                self.current_source.vars['magnitude'].view[:] = X[i] / 100.
-                self.g_model.push_var_to_device("cs",'magnitude')
+            self.current_source.vars['magnitude'].view[:] = X[i] / 100.
+            self.g_model.push_var_to_device("cs",'magnitude')
+
+            # Run simulation
+            while self.g_model.t < self.single_example_time * (i+1):
```
neworderofjamie: You could just convert ``self.single_example_time`` into timesteps by dividing by ``self.timestep`` and taking the ceiling  and then just call ``self.g_model.step_time()`` that many times

tensor_genn/algorithms/relu_ann.py
ajaysub110: I've rewritten the loop to now iterate through examples and divide each example into 'before simulation', 'during simulation' and 'after simulation'. Hope this cleans it up a bit

tnowotny:
Evaluation function looking much more solid now.

tensor_genn/algorithms/relu_ann.py
ajaysub110: Oh yes! That'll make it much more independent. I'll do that right away

neworderofjamie:
I'm happy to merge this if you are @tnowotny ?

tnowotny:
Yeah, I think this is good to go.

PR #3, opened 2019-06-26
Adds example code for convnet conversion

ajaysub110:
I've added the code for conversion of CNNs. While testing, I realized that CNNs gives a better test accuracy when the membrane capacitance is given a value of around 0.4 (which is lower than the ideal value for Dense layers = ~1.0). My guess for why this happens is that in convolution, we desire a longer integration period for a postsynaptic neuron because interdependence between input spikes is important in convolution. Does this intuition make sense or is there another reason for this observation?
I plan to add this capacitance condition directly into the library implementation, if that's alright. Please let me know if the overall conversion procedure looks okay. Thanks.


ajaysub110:
For a Tensorflow CNN test accuracy of ~98%, the converted model gets around 97%

conv.py
```diff
@@ -0,0 +1,216 @@
+import numpy as np
```
neworderofjamie: I don't really think this file needs to be part of the library

tensor_genn/algorithms/relu_ann.py
```diff
@@ -10,18 +10,67 @@ def __init__(self,neuron_resting_voltage=-60.0,neuron_threshold_voltage=-55.0,
...
+        self.Cm = membrane_capacitance # For DNN - 1.0, CNN - 0.4
         self.timestep = model_timestep
         self.single_example_time = single_example_time
 
+    def create_weight_matrices(self, tf_model):
+        tf_weights = tf_model.get_weights()
+        tf_layers = tf_model.layers
+
+        g_model_weights = []
+        n_units = [np.prod(tf_layers[0].input_shape[1:])]
```
neworderofjamie: a comment describing what this does would be great - it might be a little obtuse to those not well versed in numpy :smile:

tensor_genn/algorithms/relu_ann.py
```diff
@@ -69,37 +117,46 @@ def convert(self, tf_model):
...
+                "if"+str(i),n_units[i],if_model,if_params,if_init)
             )
 
             # Synapse
-            self.g_model.add_synapse_population(
-                "syn"+str(i-1)+str(i),"DENSE_INDIVIDUALG",genn_wrapper.NO_DELAY,
-                self.neuron_pops[i-1], self.neuron_pops[i],
-                "StaticPulse",{},{'g':tf_weights[i-1].reshape(-1)},{},{},
-                "DeltaCurr",{},{}
+            self.syn_pops.append(self.g_model.add_synapse_population(
```
neworderofjamie: It's very inefficient to represent dense layers using sparse connectivity - it doubles the memory usage and memory bandwidth required for spike processing so this needs a bit more thought.

tensor_genn/algorithms/relu_ann.py
```diff
@@ -10,18 +10,67 @@ def __init__(self,neuron_resting_voltage=-60.0,neuron_threshold_voltage=-55.0,
...
+
     def convert(self, tf_model):
         # Check model compatibility
         if not isinstance(tf_model,tf.keras.models.Sequential):
             raise NotImplementedError('Implementation for type {} models not found'.format(type(tf_model)))
         
         for layer in tf_model.layers[:-1]:
-            if not isinstance(layer,(tf.keras.layers.Dense,tf.keras.layers.Flatten)):
-                raise NotImplementedError('Only Dense and Flatten layers are supported')
+            if not isinstance(layer,(tf.keras.layers.Dense,tf.keras.layers.Flatten, tf.keras.layers.Conv2D, tf.keras.layers.AveragePooling2D)):
```
neworderofjamie: it might be nice to do:
```python
supported_layers = (tf.keras.layers.Dense,tf.keras.layers.Flatten, tf.keras.layers.Conv2D, 
                    tf.keras.layers.AveragePooling2D)
for layer in tf_model.layers[:-1]:
    if not isinstance(layer, supported_layers)):
        pass
``` 
otherwise this is quickly going to become a bit of a mess

tensor_genn/algorithms/relu_ann.py
```diff
@@ -69,37 +117,46 @@ def convert(self, tf_model):
         
         cs_init = {"magnitude":10.0}
 
+        # Fetch augmented weight matrices
+        g_weights, n_units = self.create_weight_matrices(tf_model)
+        gw_inds = [np.nonzero(gw) for gw in g_weights]
+        gw_vals = [g_weights[i][gw_inds[i]].reshape(-1) for i in range(len(g_weights))]
```
neworderofjamie: You could replace the ``reshape(-1)`` with ``flatten()`` to make your intentions somewhat clearer. Also you could maybe move this logic into the ``create_weight_matrices`` method which would perhaps be a nice way of tidying up creating sparse and dense connectivity (``gw_inds[i]`` could perhaps be ``None`` if the matrix should be dense)

tensor_genn/algorithms/relu_ann.py
```diff
@@ -10,18 +10,67 @@ def __init__(self,neuron_resting_voltage=-60.0,neuron_threshold_voltage=-55.0,
...
+        for layer in tf_layers:
+            if not isinstance(layer, tf.keras.layers.Flatten):
+                n_units.append(np.prod(layer.output_shape[1:]))
+                syn_weights = np.zeros((n_units[j-1],n_units[j]))
+
+                if isinstance(layer,tf.keras.layers.Dense):
+                    syn_weights = tf_weights[i-1]
+                    i += 1
+
+                elif isinstance(layer,tf.keras.layers.Conv2D):
```
neworderofjamie: Some more commenting on how this works would be really useful - the logic's kinda complicated and I definitely can't tell at a glance what's going on

tensor_genn/algorithms/relu_ann.py
```diff
@@ -10,18 +10,67 @@ def __init__(self,neuron_resting_voltage=-60.0,neuron_threshold_voltage=-55.0,
...
+        
+                    for n in range(int(n_units[j])): # output unit/conv number
+                        for k in range(kh): # kernel height
+                            syn_weights[((n//oc)%ow)*ic*sw + ((n//oc)//ow)*ic*iw*sh + k*ic*iw:
+                                        ((n//oc)%ow)*ic*sw + ((n//oc)//ow)*ic*iw*sh + k*ic*iw + kw*ic,
+                                        n] = tf_weights[i-1][k,:,:,n%oc].reshape((-1))
+
+                    i += 1
+
+                elif isinstance(layer,tf.keras.layers.AveragePooling2D):
```
neworderofjamie: Similarly, some more commenting on how this works would be really useful. Also, doesn't this approach come from a paper? If so it's always good to add a reference to the comments (also anywhere else this is the case)

conversion_example.py
```diff
@@ -7,27 +7,32 @@ def train_mnist():
...
-    return model, x_train*255., y_train, x_test*255., y_test
+    return model, x_train, y_train, x_test, y_test
 
 tf_model, x_train, y_train, x_test, y_test = train_mnist()
 
 # Create models
-relu_ann = ReLUANN()
-g_model = tg.convert_model(tf_model,relu_ann,x_test[:1000],y_test[:1000])
\ No newline at end of file
+relu_ann = ReLUANN(membrane_capacitance=0.4)
```
neworderofjamie: I agree about your intuition regarding the membrane capacitance but it would be really nice if the user didn't have to specify it manually here **and** surely you don't want this membrane capacitance for neurons following dense layers? Maybe ``ReLUANN`` could have a ``dense_membrane_capacitance`` and ``sparse_membrane_capacitance`` constructor argument which get applied to the two sorts of layers respectively (and default to 0.4 and 1.0)

tensor_genn/algorithms/relu_ann.py
ajaysub110: This approach has been used in many papers and isn't taken from one in particular. Could I just add a reference to the ReLU ANN paper at the top of the file?

neworderofjamie: sounds good

conversion_example.py
ajaysub110: I tried setting different capacitances for sparse and dense layers (as opposed to using a common value for all layers). There are a couple of differences resulting from this:
1. The optimal capacitance values are different (expected)
2. Changing the layers in the network has a drastic effect on the optimal capacitance values, and hence might require some amount of parameter searching. Is this acceptable or should we go back to using a single value for a network (all layers)

neworderofjamie: Interesting! How do you mean by "changing the layers"? The number of layers, their dimensions?

ajaysub110: The number of layers. Changing the layer dimensions doesn't seem to have much effect

neworderofjamie: I guess more layers will increase the variability in spike latency through the network so you would probably want a larger membrane time constant in deeper layers and therefore a larger capacitance - do the optimal capacitance values increase as you go deeper in the network?

ajaysub110: That makes sense! I'll try varying the capacitance values with depth and let you know.

neworderofjamie: It is also possible that one of the weight normalisation schemes from the literature may help with this problem as they generally aim to keep the activation within the linear region of the postsynaptic neurons

ajaysub110: Yes. That would probably solve it.
Varying capacitance values with depth produced the expected results. Also, keeping capacitance constant, I tried increasing single_example_time to 500 and threshold voltage to -56. This gives much better results for a wider range of capacitance values

neworderofjamie: Ok cool, maybe leave that capacitance alone for now then. Maybe you can start investigating weight normalization as part of getting a bigger CIFAR10 network working?

ajaysub110: Sure

tensor_genn/algorithms/relu_ann.py
```diff
@@ -118,36 +150,47 @@ def convert(self, tf_model):
...
-        gw_vals = [g_weights[i][gw_inds[i]].reshape(-1) for i in range(len(g_weights))]
+        gw_inds, gw_vals, n_units, relevant_layers = self.create_weight_matrices(tf_model)
 
         # Define model and populations
         self.g_model = genn_model.GeNNModel("float","g_model")
         self.neuron_pops = []
         self.syn_pops = []
 
-        for i in range(1,n_layers):
+        for i in range(1,len(relevant_layers)+1):
```
neworderofjamie:
```python
for i, (inds, vals, n) in enumerate(zip(gw_inds, gw_vals, n_units)):
    pass
```
is probably a bit tidier and more pythonic. Also I don't think ``relevant_layers`` is necessary - you can determine the type of connectivity to use by just testing ``inds is None``

tensor_genn/algorithms/relu_ann.py
```diff
@@ -118,36 +150,47 @@ def convert(self, tf_model):
...
+            if isinstance(relevant_layers[i-1],tf.keras.layers.Dense):
+                # Postsynaptic neuron
+                self.neuron_pops.append(self.g_model.add_neuron_population(
+                    "if"+str(i),n_units[i],if_model,dense_if_params,if_init)
+                )
+
+                # Synapse
+                self.syn_pops.append(self.g_model.add_synapse_population(
+                    "syn"+str(i-1)+str(i),"DENSE_INDIVIDUALG",genn_wrapper.NO_DELAY,
+                    self.neuron_pops[i-1],self.neuron_pops[i],
```
neworderofjamie: you could use ``self.neuron_pops[-2], self.neuron_pops[-1]`` to make this clearer

PR #4, opened 2019-07-16
Adds support for padding in conv2d

ajaysub110:
Hi,
Was finally able to come up with a way to add support for padding! It turned out to harder than expected since the padding is done on all sides of the input map, and I needed to properly index the subset of weight values that should be considered for each convolution.

### DIFF
--- a/tensor_genn/algorithms/relu_ann.py
+++ b/tensor_genn/algorithms/relu_ann.py
@@ -55,7 +55,7 @@ class ReLUANN():
                     ih,iw,ic = layer.input_shape[1:]
                     oh,ow,oc = layer.output_shape[1:]
 
-                    if layer.padding == 'valid':
+                    if layer.padding == 'valid': # no padding in layer i.e. output_size < input_size
                         # Proceed with weight mapping for regular convolution
                         for n in range(int(n_units[j])): # output unit index/conv number
                             for k in range(kh): # over kernel height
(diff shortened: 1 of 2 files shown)


tensor_genn/algorithms/relu_ann.py
```diff
@@ -42,23 +42,63 @@ def create_weight_matrices(self, tf_model):
...
+
+                    if layer.padding == 'valid':
+                        # Proceed with weight mapping for regular convolution
+                        for n in range(int(n_units[j])): # output unit index/conv number
+                            for k in range(kh): # over kernel height
+                                syn_weights[((n//oc)%ow)*ic*sw + ((n//oc)//ow)*ic*iw*sh + k*ic*iw:
+                                            ((n//oc)%ow)*ic*sw + ((n//oc)//ow)*ic*iw*sh + k*ic*iw + kw*ic,
+                                            n] = tf_weights[i-1][k,:,:,n%oc].flatten()
+
+                    elif layer.padding == 'same':
```
neworderofjamie: I might be nice to explain what 'same' padding is briefly in a comment i.e. that it keeps the output size the same as the input side

tensor_genn/algorithms/relu_ann.py
```diff
@@ -42,23 +42,63 @@ def create_weight_matrices(self, tf_model):
...
-                            '''
-                            Indexing of neurons is done in the order: channels -> width -> height
-                            Assign weights for every kw*ic block in the input and map to output neuron
-                            corresponding to that convolution operation.
-                            '''
-                            syn_weights[((n//oc)%ow)*ic*sw + ((n//oc)//ow)*ic*iw*sh + k*ic*iw:
-                                        ((n//oc)%ow)*ic*sw + ((n//oc)//ow)*ic*iw*sh + k*ic*iw + kw*ic,
-                                        n] = tf_weights[i-1][k,:,:,n%oc].flatten()
+
+                    if layer.padding == 'valid':
```
neworderofjamie: It might be nice to note in a comment here that 'valid' infact means no padding!

PR #5, opened 2019-07-27
Data-based normalization

ajaysub110:


tensor_genn/algorithms/relu_ann.py
```diff
@@ -255,7 +258,7 @@ def evaluate(self, X, y=None):
                     npop.vars["Vmem"].view[:] = random.uniform(self.Vres,self.Vthr)
                     self.g_model.push_state_to_device("if"+str(j))
                 
-            self.current_source.vars['magnitude'].view[:] = X[i] / 100.
+            self.current_source.vars['magnitude'].view[:] = X[i]*1000.
```
neworderofjamie: Where are these 'magic numbers' coming from? As discussed by email I think you could well be 'over-saturating' the input to the neurons. Assuming that X is normalised (i.e. everything is between 0 and 1), you probably want to scale by ``($(Vthr) / (DT / $(Cm)`` so an input of 1 makes the neuron spike immediately and lower levels of input result in longer time to first spike.

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self,data=None, batch_size=100):
+        self.x_train = data
+        self.batch_size = batch_size # Adjust based on available system memory
+
+    def normalize(self,tf_model):
+        # Get parameterized layer indices
+        wts_exist = [layer.get_weights() != [] for layer in tf_model.layers]
+        wts_exist_inds = [i for i, x in enumerate(wts_exist) if x == True]
+
+        # Function to fetch ReLU outputs from intermediate layers
```
neworderofjamie: ``wts_exist = [(i, layer) for layer in tf_model.layers if layer.get_weights().size > 0]`` might be slightly cleaner

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+        get_output_from_layer = K.function([tf_model.layers[0].input],
+                                           [tf_model.layers[i].output for i in wts_exist_inds])
+
+        n_batches = math.ceil(len(self.x_train)/self.batch_size)
+        max_wts = [np.max(wt) for wt in tf_model.get_weights()] # maximum input weights for each layer
+        max_acts = [0]*len(wts_exist_inds)
+        for bi in range(n_batches):
+            # get training set batch
+            if bi < n_batches - 1:
+                x = self.x_train[bi*self.batch_size:(bi+1)*self.batch_size]
```
neworderofjamie: how come this logic doesn't work for both cases?

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+            else:
+                x = self.x_train[bi*self.batch_size:]
+
+            # maximum output activations for each layer
+            layer_outputs = get_output_from_layer([x])
+            b_macts = [np.max(op) for op in layer_outputs]
+            max_acts = [max(b_macts[i],max_acts[i]) for i in range(len(b_macts))]
+        
+        # Compute scale factors and normalize weights
+        scale_factors = [max(max_acts[i],max_wts[i]) for i in range(len(max_acts))]
```
neworderofjamie: ``scale_factors = [max(a, w) for a, w in zip(max_acts, max_wts)]`` might be a little cleaner

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+            # get training set batch
+            if bi < n_batches - 1:
+                x = self.x_train[bi*self.batch_size:(bi+1)*self.batch_size]
+            else:
+                x = self.x_train[bi*self.batch_size:]
+
+            # maximum output activations for each layer
+            layer_outputs = get_output_from_layer([x])
+            b_macts = [np.max(op) for op in layer_outputs]
+            max_acts = [max(b_macts[i],max_acts[i]) for i in range(len(b_macts))]
```
neworderofjamie: ``max_acts = [max(b, a) for a, w in zip(b_macts, max_acts)]`` might be a little cleaner

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+        # Get parameterized layer indices
+        wts_exist = [layer.get_weights() != [] for layer in tf_model.layers]
+        wts_exist_inds = [i for i, x in enumerate(wts_exist) if x == True]
+
+        # Function to fetch ReLU outputs from intermediate layers
+        get_output_from_layer = K.function([tf_model.layers[0].input],
+                                           [tf_model.layers[i].output for i in wts_exist_inds])
+
+        n_batches = math.ceil(len(self.x_train)/self.batch_size)
+        max_wts = [np.max(wt) for wt in tf_model.get_weights()] # maximum input weights for each layer
```
neworderofjamie: How does the output of ``tf_model.get_weights()`` correspond to your ``wts_exist_inds``? Are the layers without weights not included?

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+        max_acts = [0]*len(wts_exist_inds)
+        for bi in range(n_batches):
```
neworderofjamie: How come this needs to operate on batches at all? Why not just loop through all of ``self.x_train``?

tensor_genn/algorithms/weight_normalization/data_based.py
```diff
@@ -0,0 +1,48 @@
...
+    def normalize(self,tf_model):
+        # Get parameterized layer indices
+        wts_exist = [layer.get_weights() != [] for layer in tf_model.layers]
+        wts_exist_inds = [i for i, x in enumerate(wts_exist) if x == True]
+
+        # Function to fetch ReLU outputs from intermediate layers
+        get_output_from_layer = K.function([tf_model.layers[0].input],
+                                           [tf_model.layers[i].output for i in wts_exist_inds])
+
+        n_batches = math.ceil(len(self.x_train)/self.batch_size)
```
neworderofjamie: It would be really great to include a longer comment here explaining the algorithm at a higher level i.e. saying that, across all the training data, you find the maximum activation of each layer, multiple this by the weights in the subsequent layer and then divide by whatever :smile:

tensor_genn/algorithms/weight_normalization/data_based.py
ajaysub110: I've split it into cases since the last batch might not contain a `batch_size` number of samples. So I only take the elements till the end, in that case. Hope that makes sense.

tensor_genn/algorithms/weight_normalization/data_based.py
ajaysub110: `tf_model.get_weights()` returns only the array of existent weights in the model while `layer.get_weights()` returns the weights of a particular layer (`[]` if it has none). So in order to get `wts_exist_inds` I access the indices of only the layers where `layer.get_weights() != []`

tensor_genn/algorithms/weight_normalization/data_based.py
ajaysub110: For large training set sizes, the operation exceeds memory. So, I thought of splitting the operations into smaller batch sizes, small enough to not overflow RAM.

neworderofjamie: Ahhh I see - makes total sense - it's nothing to do with batches in the training regime sense

tensor_genn/algorithms/weight_normalization/data_based.py
neworderofjamie: you could put an ``assert len(tf_model.get_weights()) == len(wts_exist)`` to make the assumption that they are the same explicit

ajaysub110: Sure. It will enhance readability of the code

neworderofjamie:
Did my intuition about scaling the input current help the performance?

ajaysub110:
Yes. It has improved the consistency of the algorithm over a larger range of parameter values, although the accuracy remains more or less the same on both datasets. And the required single_example_time too is around 2500 (same as before). 
I have completed the raster plot code as well. Maybe we could now move on to searching for insights from that?

PR #6, opened 2019-08-16
Raster plot

ajaysub110:


tensor_genn/algorithms/relu_ann.py
```diff
@@ -242,31 +246,62 @@ def convert(self, tf_model, scaled_tf_weights=None):
 
         return self.g_model, self.neuron_pops, self.current_source
 
-    def evaluate(self, X, y=None):
+    def evaluate(self, X, y=None, raster_plot=False):
```
neworderofjamie: I don't really like having the raster plotting functionality **in** this function - for a start it'll need to be duplicated in every other algorithm and also it doesn't allow users to do anything else with the data. Perhaps a better API would be:
```python
def  evaluate(self, X, y=None, save_example_spikes=[]):
```

tensor_genn/algorithms/relu_ann.py
```diff
@@ -242,31 +246,62 @@ def convert(self, tf_model, scaled_tf_weights=None):
...
                 
             self.current_source.vars['magnitude'].view[:] = X[i]*(-self.Vthr * (self.dCm/self.timestep))
             self.g_model.push_var_to_device("cs",'magnitude')
 
             # Run simulation
-            for _ in range(math.ceil(self.single_example_time/self.timestep)):
+            for t in range(math.ceil(self.single_example_time/self.timestep)):
                 self.g_model.step_time()
 
+                if raster_plot and i == n_examples-1:
```
neworderofjamie: This could then be replaced with:

```
if i in save_example_spikes:
```

tensor_genn/algorithms/relu_ann.py
```diff
@@ -242,31 +246,62 @@ def convert(self, tf_model, scaled_tf_weights=None):
...
+                            spike_times[j] = np.hstack((spike_times[j], ts))
+
             # After simulation
             self.g_model.pull_var_from_device("if"+str(n-1),'SpikeNumber')
             SpikeNumber_view = self.neuron_pops[-1].vars["SpikeNumber"].view
             n_correct += (np.argmax(SpikeNumber_view)==y[i])
+
         accuracy = (n_correct / n_examples) * 100.
+
+        if raster_plot:
```
neworderofjamie: You could then make a ``tensor_genn.utils.plotting`` module and have the functionality to plot your spike data structure in there.

tensor_genn/algorithms/relu_ann.py
```diff
@@ -246,15 +243,15 @@ def convert(self, tf_model, scaled_tf_weights=None):
...
         n_examples = len(X)
         X = X.reshape(n_examples,-1)
         y = y.reshape(n_examples)
         n = len(self.neuron_pops)
         
         n_correct = 0
-        spike_ids = [None]*len(self.neuron_pops)
-        spike_times = [None]*len(self.neuron_pops)
+        spike_ids = [[None]*len(self.neuron_pops)]*len(save_example_spikes)
+        spike_times = [[None]*len(self.neuron_pops)]*len(save_example_spikes)
```
neworderofjamie: sadly, this doesn't work:
```python
In [5]: x= [[None]*4]*2                                                                                                         
In [6]: x                                                                                                                       
Out[6]: [[None, None, None, None], [None, None, None, None]]
In [7]: x[0][0]=1                                                                                                               
In [8]: x                                                                                                                      
Out[8]: [[1, None, None, None], [1, None, None, None]]
```
I think the best you can do is:
```python
spike_times = [[None for _ in enumerate(self.neuron_pops))] for _ in enumerate(save_example_spikes)]     
```

tensor_genn/algorithms/relu_ann.py
```diff
@@ -269,39 +266,26 @@ def evaluate(self, X, y=None, raster_plot=False):
             for t in range(math.ceil(self.single_example_time/self.timestep)):
                 self.g_model.step_time()
 
-                if raster_plot and i == n_examples-1:
+                if i in save_example_spikes:
```
neworderofjamie: You could probably save looking up i twice by doing this in :
```python
try:
    k = save_example_spikes.index(i)
    ...
    ...
except ValueError:
    pass
```

tensor_genn/algorithms/relu_ann.py
ajaysub110: Oops. I'll change it right away

PR #21, opened 2021-03-04
size one batch fix

### LINKED ISSUE #20: view reduced to a size 1 array
I was getting low performance in the GeNN version of the simple CNN example and I think it may be due to using a batch size of 1 and the following lines of the evaluate method 
https://github.com/genn-team/ml_genn/blob/a88cf1af6bddf95e4e31c38af0d80cb12ea25e3e/ml_genn/model.py#L212-L214

I just changed it to 
```python
output_view = nrn.vars['nSpk'].view.reshape(batch_n, -1)
```
and it seems to be working ok now.

jamesturner246:
@chanokin pointed out that size one batch output was not being indexed correctly. This fixes #20.

### DIFF
--- a/ml_genn/layers/input_layer.py
+++ b/ml_genn/layers/input_layer.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from ml_genn.layers.base_layer import BaseLayer
 from ml_genn.layers.input_neurons import InputNeurons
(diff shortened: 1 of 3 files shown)


neworderofjamie (approved):
Looks good!

PR #23, opened 2021-04-06
Fixed bugs with spike recording

neworderofjamie:
In the process of trying to add FS neurons to mlGeNN, I hit some problems with recording which I tried to fix here (at some point mlGeNN should be updated to use the [new spike recording system](https://github.com/genn-team/genn/pull/372) but that's one for another day):

* Use correct syntax for accessing neuron populations
* Correctly handle batching when accessing current spikes
* Fixed classic Python bug involving empty list initialization

The classic bug to which I refer is this:
```python
x = [[]] * 4
x[0].append(5)
print(x)
```
which actually creates a list consisting of 4 references to the same (initially) empty list so the result is:
```
[[5], [5], [5], [5]]
```
standard (minimally ugly) solution which  I've applied in Model is to to do something like:
```python
x = [[] for _ in range(4)]
```

### DIFF
--- a/examples/simple_cnn.py
+++ b/examples/simple_cnn.py
@@ -58,6 +58,5 @@ if __name__ == '__main__':
     # Report ML GeNN model results
     print('Accuracy of SimpleCNN GeNN model: {}%'.format(acc[0]))
     if args.plot:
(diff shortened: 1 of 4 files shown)


jamesturner246 (approved):
Fair. I would have definitely missed that. Go ahead.

PR #25, opened 2021-07-05
hashable weakrefs

jamesturner246:
Replace all weakref.proxy with weakref.ref.


neworderofjamie (approved):
Nice! The question is...are they now hashable? 😄

jamesturner246:
They are indeed!

PR #26, opened 2021-07-07
Refactor converters

jamesturner246:
simple (no norm conversion, data-norm, spike-norm, few-spike are chosen with --converter command line option.


examples/simple_cnn.py
```diff
@@ -46,20 +46,23 @@
...
     print("TF evaluation:%f" % (perf_counter() - tf_eval_start_time))
 
-    # Create, suitable converter to convert TF model to ML GeNN
-    converter = (FewSpike(K=8, norm_data=[x_norm]) if args.few_spike 
-                 else RateBased(input_type=args.input_type, 
-                                norm_data=[x_norm],
-                                norm_method=args.norm_method,
-                                spike_norm_time=500))
+    # Create a suitable converter to convert TF model to ML GeNN
+    if args.converter == 'few-spike':
```
neworderofjamie, line 50: You could add a helper to args containing this logic to avoid copying it in all the examples. Maybe something like:
```python
def build_converter(self, norm_data, K=8, norm_time=500):
    if self.converter == 'few-spike':
        return FewSpike(K=K, norm_data=[norm_data])
    elif args.converter == 'data-norm':
        converter = DataNorm(norm_data=[norm_data], input_type=self.input_type)
    elif args.converter == 'spike-norm':
        return SpikeNorm(norm_data=[norm_data], norm_time=norm_time, input_type=self.input_type)
    else:
        return Simple(input_type=self.input_type)
```

ml_genn/converters/few_spike.py
```diff
@@ -23,19 +22,19 @@ def validate_tf_layer(self, tf_layer):
         if tf_layer.use_bias == True:
             raise NotImplementedError('bias tensors not supported')
 
-    def create_input_neurons(self, norm_output):
-        alpha = (self.alpha if norm_output.max_input is None 
-                 else float(np.ceil(norm_output.max_input)))
+    def create_input_layer(self, pre_compile_output):
```
neworderofjamie, line 25: Why this rename? It returns a neuron object not a layer object

ml_genn/converters/few_spike.py
```diff
@@ -23,19 +22,19 @@ def validate_tf_layer(self, tf_layer):
...
+        alpha = (self.alpha if pre_compile_output.max_input is None 
+                 else float(np.ceil(pre_compile_output.max_input)))
         return FSReluInputNeurons(self.K, alpha, self.signed_input)
 
-    def create_neurons(self, tf_layer, norm_output):
+    def create_layer(self, tf_layer, pre_compile_output):
```
neworderofjamie, line 30: Likewise with this rename

ml_genn/converters/few_spike.py
```diff
@@ -23,19 +22,19 @@ def validate_tf_layer(self, tf_layer):
...
         # Lookup optimised alpha value for neuron
-        alpha = (float(np.ceil(norm_output.max_activations[tf_layer]))
-                 if tf_layer in norm_output.max_activations 
+        alpha = (float(np.ceil(pre_compile_output.max_activations[tf_layer]))
+                 if tf_layer in pre_compile_output.max_activations 
                  else self.alpha)
         return FSReluNeurons(self.K, alpha)
     
-    def normalise_pre_compile(self, tf_model):
```
neworderofjamie, line 38: This is a good rename however 😄

neworderofjamie (commented):
I think that one rename doesn't make sense but, otherwise, this is definitely an improvement - nice!

jamesturner246:
Yeah, you're right about the create_* functions. I reverted that rename and did the argparse suggestion.

neworderofjamie (approved):

PR #27, opened 2021-07-12
Changes to add one hot encoding

jfgf11:


neworderofjamie (commented):
This obviously works but, I think, it would better to figure out if labels are one-hot encoded or not automatically based on whether each label is a scalar or a vector of the same size as the output layer. What do you think @jamesturner246 @tnowotny ?

jamesturner246:
@neworderofjamie , agree that would be more robust. You can access the shape of all layers, and infer that a full output size vector is one-hot, while a scalar is not. This could be done outside of the loop.

neworderofjamie:
@jfgf11 you are more than welcome to have a go with this but, if not, using the "sparse_categorical_crossentropy" loss when you train lets you train directly on the un-encoded labels,.

Lala341:
@neworderofjamie We made changes to check if each label is a scalar or a vector of the same size.

ml_genn/model.py
```diff
@@ -234,6 +234,7 @@ def evaluate(self, data, labels, time, save_samples=[]):
                 for output_i in range(len(self.outputs)):
                     predictions = self.outputs[output_i].neurons.get_predictions(
                         pipe_batch_end - pipe_batch_start)
+                    batch_labels = [np.argmax(i) for i in batch_labels] if (batch_labels[0].shape != predictions[0].shape) else batch_labels
```
neworderofjamie, line 237: It might be a little more readable (we _try_ and avoid really long lines) if you did the test on the shapes outside the loop at the top of the function like:
```python
category_labels = (batch_labels[0].shape != predictions[0].shape)
```
then you can simplify this line to:
```python
batch_labels = [np.argmax(i) for i in batch_labels] if category_labels else batch_labels
```

Lala341: Thanks, yes, it's more readable, it's in the new commit.

neworderofjamie (approved):
Great! What do you think @jamesturner246

jamesturner246 (approved):
Looks good. Thanks guys.

PR #28, opened 2021-07-15
Convert both Sequential and Funcitonal TensorFlow models

jamesturner246:
This change modifies the `Model.convert_tf_model` function, allowing the conversion of both Sequential and Functional TensorFlow models.

The same function accepts Functional models, in addition to Sequential models as before. Add layers in TF models are parsed, and the inputs to upstream add layers are accumulated in lists passed to `Layer.connect` of the ml_genn model, together with hand-crafted `Synapses` instances (of the same type as the  upstream ml_genn `Layer`s synapses). AveragePooling layers are still handled correctly. This should allow the conversion of pre-trained TF ResNet models.

A `Model.summary` method is also added, which prints a text summary of the model.


ml_genn/model.py
```diff
@@ -274,6 +273,20 @@ def get_kernel_times(self):
...
+        # layers should already be topologically sorted
+        print('===== Summary of {} ====='.format(self.name))
+
+        for l in self.layers:
+            print('\nname: {}  shape: {}  type: {}'.format(
+                l.name, l.shape, l.__class__.__name__))
+
+            if isinstance(l, Layer):
+                for s in l.upstream_synapses:
+                    print('connections: {}'.format(
```
tnowotny, line 288: If I understand it right these are the incoming connections; maybe it would be better to disambiguate by printing out "incoming connections" instead of "connections:"?

ml_genn/model.py
```diff
@@ -290,108 +303,229 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+                tf_out_layers = set(tf_out_layers_all[tf_layer])
+
+                # skip if we still need to configure inbound layers
+                if not traversed_tf_layers.issuperset(tf_in_layers):
+                    continue
+
+                # add this layer to new layers list
+                new_tf_layers.add(tf_layer)
+
+                # traverse ignored layers to find more inputs
```
tnowotny, line 414: I don't fully understand the logic of this function; it is searching through all input layers and for those of the "ignored type" it is adding their inputs?
But the layers of ignored type remain in the set as well? They are only not added if they are the inputs to already ignored layers? - It seems rather fine-grained and convoluted. Is it really necessary and the best way of doing it?

tnowotny (commented):
Overall I think it looks sensible. I am wondering whether you could add some test(s) to keep track of that this is indeed working?
Otherwise, see comments.

ml_genn/model.py
jamesturner246: This part traverses through incoming Add, Flatten, Dropout layers to find actual weighted layers to connect to. Only non-ignored (i.e. weighted and pooling) layers are added to `tf_in_layers`. It is necessary, since there maybe multiple ignored layers to traverse before we find any good layers.

ml_genn/model.py
jamesturner246: I'll change it.

jamesturner246:
I attach a little test script to show results. Just rename it to testing.py.
[testing.txt](https://github.com/genn-team/ml_genn/files/6823736/testing.txt)

ml_genn/model.py
```diff
@@ -290,108 +303,229 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+
+            mlg_layer_lookup[tf_layer] = mlg_layer
+            mlg_model_inputs.append(mlg_layer)
+
+        # while there are still layers to traverse
+        while new_tf_layers:
+            new_tf_layer = new_tf_layers.pop()
+            traversed_tf_layers.add(new_tf_layer)
+
+            # get next TF layer to configure
```
neworderofjamie, line 402: Why not follow more of a classic breadth-first-search/traversal:
1. Add the inputs to ``new_tf_layers`` outside the while loop
2. In the while loop, first of all, handle processing of all layer types including input
3. Add outputs to the ``new_tf_layers``

I may be missing something, but it would make the algorithm a little clearer to me

ml_genn/model.py
```diff
@@ -290,108 +303,229 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+
+                        else:
+                            synapses.append(DenseSynapses(units=tf_layer.units))
+
+                    # connect layer and set weights
+                    mlg_layer.connect(sources, synapses)
+                    mlg_layer.set_weights(weights)
+
+                    mlg_layer_lookup[tf_layer] = mlg_layer
+                    if len(tf_out_layers) == 0:
```
neworderofjamie, line 460: Why does this make this layer an output?

ml_genn/converters/few_spike.py
```diff
@@ -67,4 +67,7 @@ def pre_compile(self, tf_model):
             return PreCompileOutput(max_activations={}, max_input=None)
     
     def post_compile(self, mlg_model):
-        pass
+        # do not allow multiple input or output layers
+        if len(mlg_model.inputs) > 1 or len(mlg_model.outputs) > 1:
```
neworderofjamie, line 71: why are either of things problematic for few-spike? without adding synaptic delays (which @tnowotny agrees is the solution), issues are going to occur if the graph contains branching paths with different numbers of ml genn layers which subsequently rejoin (if that makes sense)

ml_genn/model.py
jamesturner246: If no layer outputs, then it is probably an output layer. I could change it to accept a list of the tf_layers you want as outputs, but this seemed a reasonable default.

ml_genn/model.py
```diff
@@ -290,108 +303,229 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+        mlg_layer_lookup = {}
+        new_tf_layers = set()
+        traversed_tf_layers = set()
+
+        if isinstance(tf_model, tf.keras.models.Sequential):
+            tf_in_layers = tf_in_layers_all[tf_model.layers[0]]
+            assert(len(tf_in_layers) == 1)
+            tf_out_layers = [n.outbound_layer for n in tf_in_layers[0].outbound_nodes
+                             if n in tf_model_nodes]
+            tf_out_layers_all[tf_in_layers[0]] = tf_out_layers
```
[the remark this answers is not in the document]
jamesturner246: Because Sequential models do not contain the InputLayer instance, so traversing `seq_model.layers` does not find the input layer, and we need to find output layers for *all* tf layers, including the input.

ml_genn/model.py
```diff
@@ -290,108 +303,229 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+
+        else:
+            tf_in_layers = [tf_model.get_layer(name) for name in tf_model.input_names]
```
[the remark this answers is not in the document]
jamesturner246: No, because Sequential models get rid of their input layer object. See below.

ml_genn/model.py
jamesturner246: I need *all* upstream layers to be created before I create the new one. Breadth first search may not guarantee this in all cases.

ml_genn/model.py
```diff
@@ -290,108 +303,229 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+                    tf_layer.__class__.__name__))
+            elif isinstance(tf_layer, weighted_tf_layers):
                 converter.validate_tf_layer(tf_layer)
 
-        # Perform any pre-compilation tasks
-        pre_compile_output = converter.pre_compile(tf_model)
-
-        model = Model(name=tf_model.name)
+        # only traverse nodes belonging to this model
+        tf_model_nodes = []
```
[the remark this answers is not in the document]
jamesturner246: TensorFlow stopped supporting python 2 at version 2.1.

ml_genn/converters/few_spike.py
jamesturner246: Eventually, yes. Just wanted to get the basic version out now :)

ml_genn/model.py
tnowotny: Ok, thanks ... that makes sense but does it matter that the original layers of ignored type stay in the list ```tf_in_layers```? Is that intentional?

ml_genn/model.py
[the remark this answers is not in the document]
neworderofjamie: fair enough - no need to mess about with six then

ml_genn/model.py
jamesturner246: They are deleted in line 420: `tf_in_layers.discard ...`.

jamesturner246: I just checked the code, and no ignored layers are present in `tf_in_layers` after this point. If you have some counterexample code, I would be very interested!

ml_genn/model.py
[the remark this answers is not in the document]
neworderofjamie: Ahh I see - ``input_names`` has it's name but for whatever reason it can't be extracted - bizarre. Maybe add a  comment about why you handle the two differently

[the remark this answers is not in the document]
jamesturner246: Allright, will do.

ml_genn/model.py
neworderofjamie: I'm guessing those cases come from when you have multiple inputs?

ml_genn/model.py
tnowotny: Ok ... I see now ... it's ok. I find the logic a bit convoluted tbh 
It would be slightly more straightforward if you started with an empty set
```final_tf_in_layers``` and then added into it whenever a layer is "not ignored" rather than adding and discarding ... a bit like this:
```
 final_tf_in_layers = set()
     while tf_in_layers:
         tf_in_layer = tf_in_layers.pop()
         if isinstance(tf_in_layer, ignored_tf_layers):
             tf_in_layers.update(tf_in_layers_all[tf_in_layer])
         else:
             final_tf_in_layers.add(tf_in_layer)
tf_in_layers= final_tf_in_layers```

ml_genn/model.py
jamesturner246: For instance.

ml_genn/model.py
neworderofjamie: ahh ok - that makes sense - a comment to explain might be nice. It's really annoying that you can output tensors at the TF level with ``tf_model.outputs`` but, as far as I can see, there's no way to go from that to the keras level layer object.

ml_genn/model.py
jamesturner246: Done.

ml_genn/model.py
jamesturner246: There's a lot of trivial graph traversal things you probably ought to be able to do, but can't without ugly hacking, in Keras... :(

jamesturner246:
Okay, I've done all the suggested changes from the comments I've 'resolved' above. I've also added tests for Sequential and Functional model conversion in ./tests/test_tf_conversion.py. Let me know if there's anything else. Cheers!

neworderofjamie (approved):

jamesturner246:
Huzzah! Thank you both.

Issue #30, opened 2021-07-22
GlobalAveragePooling2D  layers


neworderofjamie:
These can be implemented using average pooling layers but the pool is the entire previous layer (so they pool a whole channel into one value). See https://keras.io/api/layers/pooling_layers/global_average_pooling2d/ and https://machinelearningmastery.com/pooling-layers-for-convolutional-neural-networks/

jamesturner246:
This is a trivial one. I'll tweak the conversion code to handle globalaveragepooling layers. I'm not sure it's worth adding a new ml_genn layer, but I could do if it's really wanted.

neworderofjamie:
I think tweaking the conversation code is the way to go

jamesturner246:
Done in #31

PR #31, opened 2021-07-22
Add support for TF GlobalAveragePooling2D layers in conversion code.

jamesturner246:
Adds TF GlobalAveragePooling2D conversion code, fixing #30.

### DIFF
--- a/ml_genn/layers/avepool2d_conv2d_synapses.py
+++ b/ml_genn/layers/avepool2d_conv2d_synapses.py
@@ -95,14 +95,24 @@ class AvePool2DConv2DSynapses(BaseSynapses):
                  conv_padding='valid', connectivity_type='procedural'):
         super(AvePool2DConv2DSynapses, self).__init__()
         self.filters = filters
-        self.pool_size = pool_size
-        self.conv_size = conv_size
-        if pool_strides == None:
+        if isinstance(pool_size, int):
+            self.pool_size = (pool_size, pool_size)
+        else:
+            self.pool_size = pool_size
+        if isinstance(conv_size, int):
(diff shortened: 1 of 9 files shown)


neworderofjamie (commented):
maybe a test would be good?

jamesturner246:
Made some changes and added a couple of tests. Basically solved a silly mistake, allow pool/conv size/stride to be passed as an integer or a tuple, and removed the code to have conv2d layers after global avepoo2d layers, since this doesn't make sense.

ml_genn/layers/avepool2d_conv2d_synapses.py
```diff
@@ -95,14 +95,24 @@ def __init__(self, filters, pool_size, conv_size, pool_strides=None,
...
         super(AvePool2DConv2DSynapses, self).__init__()
         self.filters = filters
-        self.pool_size = pool_size
-        self.conv_size = conv_size
-        if pool_strides == None:
-            self.pool_strides = (pool_size[0], pool_size[1])
+        if isinstance(pool_size, int):
+            self.pool_size = (pool_size, pool_size)
+        else:
+            self.pool_size = pool_size
```
neworderofjamie, line 101: this could perhaps go in a helper function as I suspect it's fairly ubiquitous 
```python
def getParam2D(param, default=None):
    if param is None:
        if default is not None:
            return default
        else:
            raise Exception("Parameter required")
    elif isinstance(param, int):
        return (param, param)
    else:
        return param
```

jamesturner246: Done.

PR #33, opened 2021-07-27
Pool multi input

jamesturner246:
This change allows TF models containing pool layers with multiple inputs (via Add layers) to be converted into MLG models. This will enable more types of ResNet models to be converted - in particular, ResNet models with global average pooling after skip junctions.

Tests have also been added.

### DIFF
--- a/ml_genn/model.py
+++ b/ml_genn/model.py
@@ -455,7 +455,7 @@ class Model(object):
                             while pool_in_layers:
                                 pool_in_layer = pool_in_layers.pop()
                                 if isinstance(pool_in_layer, ignored_tf_layers):
-                                    pool_in_layers.update(tf_in_layers_all[tf_pool_in_layer])
+                                    pool_in_layers.update(tf_in_layers_all[pool_in_layer])
                                 else:
                                     final_pool_in_layers.add(pool_in_layer)
                             pool_in_layers = final_pool_in_layers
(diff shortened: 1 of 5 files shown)


ml_genn/model.py
```diff
@@ -434,35 +434,57 @@ def convert_tf_model(tf_model, converter=Simple(),
...
+                            # traverse ignored layers to find more pool layer inputs
+                            pool_in_layers = set(tf_in_layers_all[tf_in_layer])
+                            final_pool_in_layers = set()
+                            while pool_in_layers:
+                                pool_in_layer = pool_in_layers.pop()
+                                if isinstance(pool_in_layer, ignored_tf_layers):
+                                    pool_in_layers.update(tf_in_layers_all[pool_in_layer])
+                                else:
+                                    final_pool_in_layers.add(pool_in_layer)
+                            pool_in_layers = final_pool_in_layers
```
neworderofjamie, line 461: This logic is duplicated in two other places ([here](https://github.com/genn-team/ml_genn/pull/33/files#diff-1c228c3438adda0d7b0fef132e5b9a1e9b281ef7b7d70f810e08ad4477e8126eR421-R428) and [here](https://github.com/genn-team/ml_genn/pull/33/files#diff-1c228c3438adda0d7b0fef132e5b9a1e9b281ef7b7d70f810e08ad4477e8126eR515-R522)). Maybe this could be moved out into a helper function to aid the readability of this

neworderofjamie (approved):
This is becoming a bit of an overly large function - I've made one suggestion for one bit of functionality to move out but maybe, next bit of functionality you add here, it might be worth refactoring a little more

ml_genn/model.py
jamesturner246: Done.

Issue #29, opened 2021-07-22
Activation layers


neworderofjamie:
As well as specifying the activation via the ``activation="xxx"`` kwarg of layer objects, you can also specify it with specific activation layers. See https://keras.io/api/layers/activations/

jamesturner246:
This would need to be limited to after non-activation layers, such as pooling layers. It could be done in a manner similar to `AvePooling2DDense`, but with identity Dense weights. So MLG might convert a TF model pooling layer -> activation layer into an `AvePooling2D` layer, which must have an activation function (taken from the activation layer).

The identity Dense weights would not need to be stored - it could all be done in procedural connectivity.

neworderofjamie:
To be clear, am I right in thinking that it's an alternative way of specifying the activation functions that follow layers rather than a way of letting you have multiple back-to-back activation functions (I don't really know what that would do)? If so, I totally agree it should be handled like ``AvePooling2DDense`` style stuff.

jamesturner246:
The former, I'm sure. Back to back activation seems pretty weird and pointless!

PR #34, opened 2021-08-10
Tf activation layers

jamesturner246:
This change enables the conversion of TF models with activation layers, back-to-back activation, and activation after pooling layers.

Solves #29.

To facilitate this, the change introduces two new synapse types: `IdentitySynapses` and `AvePool2DSynapses`. The`AvePool2DSynapse` synapse classes are implemented similarly to the existing `AvePool2DDenseSynapses` and `AvePool2DConv2DSynapses` classes, using procedural connectivity to conserve memory. Also the accompanying `Identity` and `AvePool2D` helper layers are included.

The model conversion process is now two-phase:
1) check the validity of the model, and construct an internal graph of configure nodes for each TF layer, which are appended to `config_steps` in order of conversion. 
2) iterare over `build_steps`, constructing layers and synapses, so long as each config node indicates that it's activation is known (thus it can be converted to an MLG layer).

Tests are included in ./tests/test_avepool2d.py and ./tests/test_identity.py.

### DIFF
--- a/ml_genn/model.py
+++ b/ml_genn/model.py
@@ -28,9 +28,11 @@ from ml_genn.converters import Simple
 from ml_genn.layers import InputLayer
 from ml_genn.layers import Layer
 
+from ml_genn.layers import IdentitySynapses
 from ml_genn.layers import DenseSynapses
-from ml_genn.layers import AvePool2DDenseSynapses
 from ml_genn.layers import Conv2DSynapses
+from ml_genn.layers import AvePool2DSynapses
+from ml_genn.layers import AvePool2DDenseSynapses
 from ml_genn.layers import AvePool2DConv2DSynapses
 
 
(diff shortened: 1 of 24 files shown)


ml_genn/model.py
```diff
@@ -356,45 +338,30 @@ def convert_tf_model(tf_model, converter=Simple(),
...
         # Perform any pre-compilation tasks
         pre_compile_output = converter.pre_compile(tf_model)
 
-        # configure ML GeNN model build process
-        mlg_model_inputs = []
-        mlg_model_outputs = []
-        mlg_layer_lookup = {}
+        # configure model build process
+        class LayerConfig(object):
+            def __init__(self):
```
neworderofjamie, line 347: why not have constructor arguments for ``is_input``, ``has_activation``, ``name``, ``shape`` and ``neurons`` (and defaults if any are ever optional) so you can go e.g.:
```python
config = LayerConfig(is_input=True, has_activation=True, name=tf_layer.name, shape=tf_layer.output_shape[0][1:], 
                     neurons=converter.create_input_neurons(pre_compile_output))
```

ml_genn/model.py
```diff
@@ -408,175 +375,319 @@ def traverse_tf_out_layers(out_layers):
...
+
+                    # configure synapses
+                    for in_config in in_configs:
+
+                        if in_config.has_activation:
+                            # configure Dense synapses
+                            config.synapses_class.append(DenseSynapses)
+                            config.synapses_params.append({'units': tf_layer.units})
+                            config.synapses_source.append(in_config)
+                            config.synapses_weights.append(tf_layer.get_weights()[0])
```
neworderofjamie, line 466: having to append stuff to four lists to add a synapse seems a bit ugly, maybe replace with a single list of ``collection.namedtuple`` e.g.
```python
InSynConfig = namedtuple('InSynConfig', ['class', 'params', 'source', 'weights'])
```
Then, this could be replaced by:
```python
config.synapses.append(InSynConfig(class=DenseSynapse, params={'units': tf_layer.units}, 
                                   source=in_config, weights=tf_layer.get_weights()[0]))
```

ml_genn/model.py
```diff
@@ -408,175 +375,319 @@ def traverse_tf_out_layers(out_layers):
...
+                # build layer
+                mlg_layer = Layer(config.name, config.neurons)
+
+                # build synapses
+                sources = []
+                synapses = []
+                weights = config.synapses_weights
+                for i in range(len(config.synapses_class)):
+                    sources.append(mlg_layer_lookup[config.synapses_source[i]])
+                    synapses.append(config.synapses_class[i](**config.synapses_params[i]))
```
neworderofjamie, line 682: with my suggested change, this would also be reduced to:
```python
sources = [mlg_layer_lookup[s.source]) for s in config.synapses]
synapses= [s.class(**s.params) for s in config.synapses]
```
which is much more pythonic

neworderofjamie (commented):
This is a big improvement and I think also a good step towards detaching the code outside the conversion function from TF.

neworderofjamie (approved):

PR #35, opened 2021-08-11
adding the one hot encoding to the model script now tested

jfgf11:
This is the one hot encoding that we made some time ago tested and fixed for all scenarios of tests that you already have and the new presented.

### DIFF
--- a/ml_genn/model.py
+++ b/ml_genn/model.py
@@ -237,6 +237,7 @@ class Model(object):
                 for output_i in range(len(self.outputs)):
                     predictions = self.outputs[output_i].neurons.get_predictions(
                         pipe_batch_end - pipe_batch_start)
+                    batch_labels[output_i] = [np.argmax(i) for i in batch_labels[output_i]] if batch_labels[output_i].shape != predictions.shape else batch_labels[output_i]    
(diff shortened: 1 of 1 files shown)


ml_genn/model.py
```diff
@@ -237,6 +237,7 @@ def evaluate(self, data, labels, time, save_samples=[]):
                 for output_i in range(len(self.outputs)):
                     predictions = self.outputs[output_i].neurons.get_predictions(
                         pipe_batch_end - pipe_batch_start)
+                    batch_labels[output_i] = [np.argmax(i) for i in batch_labels[output_i]] if batch_labels[output_i].shape != predictions.shape else batch_labels[output_i]
```
neworderofjamie, line 240: Looks good! My only comment would be that this is an overly long and overly complex line of code. This would be much clearer:
```python
if batch_labels[output_i].shape != predictions.shape:
     batch_labels[output_i] = [np.argmax(i) for i in batch_labels[output_i]]
```

neworderofjamie:
What do you think @jamesturner246? Looks good to me

neworderofjamie (approved):

jamesturner246 (approved):
Looks better. Thanks guys.

PR #36, opened 2021-08-25
Signed spike refactor

neworderofjamie:
Signed spikeness is a feature of how you train the model i.e. do you  substract means or not
* Removed signed variants from ``InputType``
* Make all converters take ``signed_input`` as constructor parameter and use this to configure input models
* ``arguments.build_converter`` takes ``signed_input`` flag as this should be set manually by model (like an optimised K)


jamesturner246 (approved):
Looks fine and straightforward.

PR #37, opened 2021-08-31
Correct few-spike conversion of branching networks

neworderofjamie:
Sorry for the chain of PRs but I wanted to build on some previous stuff. This PR fixes three issues with few-spike conversion of branching networks e.g. ResNets:

1. (Axonal) delays need to be inserted to 'balance' branches. I use the approach described [here](https://stackoverflow.com/questions/59656961/in-a-dag-how-to-find-vertices-where-paths-converge) to find where branches split and rejoin. Because the graph is a DAG, the thrilling named dominator graph can be trivially calculated directly from the [definition ](https://en.wikipedia.org/wiki/Dominator_(graph_theory)). Between each split and rejoin point, we then find all the branches and insert delays to make them all take the same time.
2. Alpha values need to be the same for all upstream populations connecting to another population. As, after 1, we know all the 'rejoin' points in the graph we can simply take the max of alpha values here.
3. The pipeline depth calculation required to read out the output at the correct time and run the network for long enough was broken. After 1, delays through all paths in the network are balanced so, assuming a single input and output, this is as simple as following an arbitrary path through the downstream synapses until you hit the output and summing the delays.

I started fiddling in test_pipeline_depth.py so ignore that for now - will replace it with some useful tests 😄

### DIFF
--- a/ml_genn/converters/data_norm.py
+++ b/ml_genn/converters/data_norm.py
@@ -63,7 +63,7 @@ class DataNorm(object):
                 'Data-Norm converter: {} layers are not supported'.format(
                     tf_layer.__class__.__name__))
 
-    def create_input_neurons(self, pre_compile_output):
+    def create_input_neurons(self, pre_convert_output):
         if self.input_type == InputType.SPIKE:
             return SpikeInputNeurons(signed_spikes=self.signed_input)
         elif self.input_type == InputType.POISSON:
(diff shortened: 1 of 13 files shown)


jamesturner246 (approved):
From what I can see, it looks good. Sequential stuff still works fine. Go ahead.

PR #42, opened 2022-01-25
Conv transpose

jamesturner246:
PR contains transpose convolution layer/synapse class. 

Also parameterises pytest tests.

Also solves bugs in pooling layers where `poolOutRow`/`poolOutCol` can exceed `pool_oh`/`pool_ow` with certain pool sizes.


Jenkinsfile
```diff
@@ -112,7 +112,7 @@ for (b = 0; b < builderNodes.size(); b++) {
                 // Checkout GeNN
                 echo "Checking out GeNN";
                 sh "rm -rf genn";
-                sh "git clone --branch toeplitz_kernelg https://github.com/genn-team/genn.git";
+                sh "git clone --branch master https://github.com/genn-team/genn.git";
```
neworderofjamie, line 115: good spot - sorry I left this is such a shit state before Christmas...

ml_genn/layers/avepool2d_conv2d_synapses.py
```diff
@@ -48,31 +48,35 @@
...
-                    // Calculate postsynaptic index and add synapse
-                    const int idPost = ((convOutRow * conv_ow * conv_oc) +
-                                        (convOutCol * conv_oc) +
-                                        outChan);
-                    $(addSynapse, idPost, kernRow, kernCol, poolInChan, outChan);
+        const int conv_ih = $(conv_ih), conv_iw = $(conv_iw), conv_ic = $(conv_ic);
+
+        if ((poolOutRow < conv_ih) && (poolOutCol < conv_iw)) {
+            // Calculate range of output rows and columns which this pool output connects to
+            const int minOutRow = min(conv_oh, max(0, 1 + ((poolOutRow + conv_padh - conv_kh) / conv_sh)));
```
neworderofjamie, line 61: These min calculations should be done like:
```c++
min((int)$(conv_oh), max(0, 1 + (int)floor((inRow + $(conv_padh) - $(conv_kh)) / $(conv_sh))))
```
to deal with the rounding issue with stride described in https://github.com/genn-team/genn/pull/489 (again, sorry for not doing this)

tests/test_avepool2d_conv2d.py
```diff
@@ -1,398 +1,72 @@
...
 import ml_genn as mlg
-from converter import Converter
 import pytest
+from converter import Converter
+
+@pytest.mark.parametrize(
+    'in_size, in_chan, out_chan, pool_size, pool_strides, conv_size, conv_strides, conv_padding, connect', [
+        (20, 1, 1, 2, 2, 3, 1, 'valid', 'sparse'),
+        (20, 1, 1, 2, 2, 3, 1, 'valid', 'procedural'),
+        #(20, 1, 1, 2, 2, 3, 1, 'valid', 'toeplitz'),
```
neworderofjamie, line 11: Is something broken with the toeplitz implementation of these?

tests/test_avepool2d_dense.py
```diff
@@ -1,241 +1,61 @@
...
-def test_avepool2d_dense_in_chan_1_padding_valid():
-    '''
-    Test AvePool2DDense with 1 input channel, 1 output channel and valid pool padding.
-    '''
-
+@pytest.mark.parametrize(
+    'in_size, in_chan, out_size, pool_size, pool_strides, connect', [
+        (20, 1, 10, 2, 2, 'sparse'),
+        (20, 1, 10, 2, 2, 'procedural'),
+        (20, 1, 10, 2, 2, 'toeplitz'),
```
neworderofjamie, line 11: There isn't a toeplitz implementation of this (as it's basically dense) so probably no point in essentially testing procedural twice

neworderofjamie (commented):
Aside from the rounding issue, this looks great. Sorry so much of it is tidying up the mess I left behind...

jamesturner246:
Thanks for that Jamie. 
I'll fix the min calculations. 

Oh yes, almost forgot. The toeplitz pool_conv was throwing a weird segfault. I will investigate this further after.

tests/test_avepool2d_dense.py
jamesturner246: Done and added error throw if one tries to use toeplitz with avepool_dense synapses.

neworderofjamie:
Ooh yeah, on the subject of loose ends, try with https://github.com/genn-team/genn/tree/missing_pygenn_toeplitz 😄

neworderofjamie:
I don't think it should be an error case - you generally want to use
toeplitz where possible as it's the fastest but not everything is supported
so it should fall back to procedural

jamesturner246:
Alright. I will change it to fall back to procedural.

Okay, that new branch seems to fix the cryptic abort issue.

It looks like the genn implementation of toeplitz avepoll_conv will still need a `poolOutCol < pool_ow` and `poolOutRow < pool_oh` check, like [here](https://github.com/genn-team/ml_genn/pull/42/files#diff-ad79adacdd41a9ed1f6388aa1134c61922f5790c3c2f73b15bae2ab9139c4d3fR59).
I'll do a quick PR off the pygenn_missing_toeplitz branch.

Also, maybe a separate issue, but I note that I get weird cryptic aborts when I use predefined constants, like `$(conv_sh)` instead of `conv_sh`. Doesn't stop me doing the minOutRow/Col fix with the `$()` version though, so maybe not important.

examples/resnet20_cifar10.py
```diff
@@ -2,17 +2,19 @@
...
+from transforms import *
 
 
 batch_size = 256
 epochs = 200
 learning_rate = 0.05
 momentum = 0.9
 
-steps_per_epoch = 50000 // epochs
+steps_per_epoch = 50000 // batch_size
```
neworderofjamie, line 17: is this right?

examples/resnet34_imagenet.py
```diff
@@ -0,0 +1,295 @@
...
+
+    # convert to float **NOTE** this converts [0,255] to [0,1) range
+    image = tf.image.convert_image_dtype(image, tf.float32)
+
+    return image, label
+
+
+#if __name__ == '__main__':
+with tf.device('/CPU:0'):
+    for gpu in tf.config.experimental.list_physical_devices('GPU'):
```
neworderofjamie, line 122: I also have a vague memory that this will crash if you've already forced the device

examples/vgg16_cifar10.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # Prepare data for ML GeNN
+    if args.n_test_samples is None:
+        args.n_test_samples = 10000
+    mlg_norm_ds = train_ds.take(args.n_norm_samples).as_numpy_iterator()
+    norm_x = np.array([d[0] for d in mlg_norm_ds])
+    mlg_validate_ds = tf.data.Dataset.from_tensor_slices((validate_x, validate_y))
+    mlg_validate_ds = mlg_validate_ds.map(color_normalize_fn, num_parallel_calls=tf.data.AUTOTUNE)
+    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: (x, y[0]), num_parallel_calls=tf.data.AUTOTUNE)
+    #mlg_validate_ds = mlg_validate_ds.take(args.n_test_samples)
```
neworderofjamie, line 153: Maybe delete all this commented out stuff

ml_genn/converters/few_spike.py
```diff
@@ -83,35 +86,52 @@ def create_input_neurons(self, pre_convert_output):
...
-            # Build dictionary of maximum activation in each layer
-            layer_alpha = {l: np.max(out) / (1.0 - 2.0 ** -self.K)
-                           for l, out in zip(weighted_layers, outputs)}
-
-            # Use input data range to directly set maximum input
-            if self.signed_input:
-                input_alpha = np.amax(np.abs(self.norm_data)) / (1.0 - 2.0 ** (1 - self.K))
+                tf_model.inputs, [l.output for l in tf_model.layers])
+
+            if isinstance(self.norm_data[0], Iterator):
```
neworderofjamie, line 108: On my 64GB(!) machine there wasn't enough memory to call ``get_outputs`` on all the dataset hence my implementing this code path. I wonder if we should also implement this for data-norm and make sure the ImageNet examples use it

neworderofjamie (changes requested):
We should also make a dependency on pygenn 4.7.0 (which I'm currently preparing for release) in the setup.py

examples/resnet20_cifar10.py
jamesturner246: Hmm. No.. Not sure how that got in there. Will change it back.

examples/resnet34_imagenet.py
```diff
@@ -0,0 +1,295 @@
...
+    image = tf.image.decode_jpeg(tf.reshape(parsed["image/encoded"], shape=[]), channels=3)
+
+    # convert to float **NOTE** this converts [0,255] to [0,1) range
+    image = tf.image.convert_image_dtype(image, tf.float32)
+
+    return image, label
+
+
+#if __name__ == '__main__':
+with tf.device('/CPU:0'):
```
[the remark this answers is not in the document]
jamesturner246: We could perhaps have separate scripts for training and validating the imagenet examples. Looking around, I'm seeing lots of people having problems releasing TF GPU memory within a process, and so far their solutions have been to just lauch a new process.

ml_genn/converters/few_spike.py
jamesturner246: I wonder if it's worth just doing batch only. Same for `evaluate` - i.e. get rid of `evaluate` and only use `evaluate_iterator`.

jamesturner246:
I think that's all of it. Sorry for the delay. Happy to chat about the changes in person.

PR #43, opened 2022-03-11
Enable FewSpike for multiple output layers. evaluate_batched now works for multi in/out layers.

jamesturner246:
Modified to allow multi-output FewSpike.

`calc_pipeline_depth` is removed, and FS layers now hold `pipeline_depth` attributes (delay_at_layer - 1). Total pipeline depth at an output layer is that layer's `pipeline_depth` attribute. Added separate label pipeline queues for each output, so each label is timed correctly.

`evaluate_batched` now handles multiple input /output layers correctly. Passed dataset should generate  `( ( in_layer_1, ..., ), ( out_layer_1, ...  ) )` tuples of tuples.

Index 0 is tuple of input layer data batches `in_layer_i.shape[0] == batch_size`, etc, and index 1 is a tuple of batched output layer labels `out_layer_j.shape[0] == batch_size` for all i and j.

### DIFF
--- a/examples/resnet20_cifar10.py
+++ b/examples/resnet20_cifar10.py
@@ -115,7 +115,7 @@ if __name__ == '__main__':
     if args.n_test_samples is None:
         args.n_test_samples = 10000
     mlg_validate_ds = validate_ds.take(args.n_test_samples)
-    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: (x, y[0]), num_parallel_calls=tf.data.AUTOTUNE)
+    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: ((x,), (y[0],)), num_parallel_calls=tf.data.AUTOTUNE)
     mlg_validate_ds = mlg_validate_ds.batch(args.batch_size)
     mlg_validate_ds = mlg_validate_ds.prefetch(tf.data.AUTOTUNE)
 
--- a/examples/resnet20_cifar10.py
+++ b/examples/resnet20_cifar10.py
@@ -115,7 +115,7 @@ if __name__ == '__main__':
     if args.n_test_samples is None:
         args.n_test_samples = 10000
     mlg_validate_ds = validate_ds.take(args.n_test_samples)
-    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: ((x,), (y[0],)), num_parallel_calls=tf.data.AUTOTUNE)
+    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: (x, y[0]), num_parallel_calls=tf.data.AUTOTUNE)
     mlg_validate_ds = mlg_validate_ds.batch(args.batch_size)
     mlg_validate_ds = mlg_validate_ds.prefetch(tf.data.AUTOTUNE)
 
--- a/ml_genn/model.py
+++ b/ml_genn/model.py
@@ -212,10 +212,10 @@ class Model(object):
 
             if data_remaining:
                 batch_x = [np.asarray(d) for d in batch_x]
-                batch_x_size = batch_x[0].shape[0]
-                batch_x_start = batch_i * self.g_model.batch_size
-                batch_x_end = batch_x_start + batch_x_size
-                save_samples_in_batch = [i for i in save_samples if batch_x_start <= i < batch_x_end]
+                batch_size = batch_x[0].shape[0]
+                batch_start = batch_i * self.g_model.batch_size
+                batch_end = batch_start + batch_size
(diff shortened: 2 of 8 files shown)


ml_genn/model.py
```diff
@@ -167,7 +167,7 @@ def evaluate_batched(self, data, time, save_samples=[]):
         """Evaluate the accuracy of a GeNN model
 
         Args:
-        data          --  an (x, y) batch dataset
+        data          --  an [[in_1, ..., in_n], [out_1, ..., out_n]] batch dataset
```
neworderofjamie, line 170: maybe you should use e.g. ``m`` and ``n`` here rather than ``n`` for both and clarify that ``m`` is the number of inputs in the model and ``n`` is the number of outputs

ml_genn/model.py
```diff
@@ -179,35 +179,35 @@ def evaluate_batched(self, data, time, save_samples=[]):
...
         batch_data, batch_labels = next(data_iter)
         batch_i = 0
 
-        batch_labels_queue = deque()
-
-        # Pad number of samples so pipeline can be flushed
-        pipeline_depth = self.calc_pipeline_depth()
-        pipeline_flush = False
-        pipeline_remaining = pipeline_depth
+        # Pipeline queues
```
neworderofjamie, line 192: It might be worth clarifying either in the comment or the variable names that these are per-output depths and queues - I didn't _quite_ get it first time I read through

ml_genn/model.py
```diff
@@ -179,35 +179,35 @@ def evaluate_batched(self, data, time, save_samples=[]):
         spike_t       --  list of spike times for each sample index in save_samples
         """
 
+        n_complete = [0] * len(self.outputs)
```
neworderofjamie, line 182: I think it would be good to check that the number of inputs and outputs in the model match the data passed to the function

ml_genn/model.py
```diff
@@ -179,35 +179,35 @@ def evaluate_batched(self, data, time, save_samples=[]):
...
 
+        # Process batches
         progress = tqdm()
-        while not pipeline_flush or pipeline_remaining > 0:
+        while data_remaining or any(len(q) > 0 for q in pipeline_labels_queue):
 
-            if not pipeline_flush:
-                batch_data = [np.array(batch_data)]
+            if data_remaining:
+                batch_data = [np.array(d) for d in batch_data]
```
neworderofjamie, line 201: I think it would be better to use ``np.asarray`` here as that won't copy if the data is already in a numpy array

ml_genn/model.py
```diff
@@ -305,23 +305,23 @@ def evaluate(self, data, labels, time, save_samples=[]):
...
+        pipeline_depth = [l.pipeline_depth if hasattr(l, 'pipeline_depth') else 0 for l in self.outputs]
+        padded_n_samples = [n_samples + (depth * self.g_model.batch_size) for depth in pipeline_depth]
 
         # Process batches
         progress = tqdm(total=n_samples)
-        for batch_start in range(0, padded_n_samples, self.g_model.batch_size):
+        for batch_start in range(0, max(padded_n_samples), self.g_model.batch_size):
 
             # If any elements of this batch have data (rather than being entirely pipeline padding)
             if batch_start < n_samples:
```
neworderofjamie, line 322: The body of the loops here and in ``evaluate_batched`` are very similar so I wonder if they could be moved into a shared helper method - I _think_ you could pass in the next batch of data for each input (if there is one) and label for each output (if there is one) and the rest could be pretty generic (although I may be missing something)

examples/resnet20_cifar10.py
```diff
@@ -115,7 +115,7 @@ def resnet20():
     if args.n_test_samples is None:
         args.n_test_samples = 10000
     mlg_validate_ds = validate_ds.take(args.n_test_samples)
-    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: (x, y[0]), num_parallel_calls=tf.data.AUTOTUNE)
+    mlg_validate_ds = mlg_validate_ds.map(lambda x, y: ((x,), (y[0],)), num_parallel_calls=tf.data.AUTOTUNE)
```
neworderofjamie, line 118: This is kinda ugly. In the very common case where there's only one input and output, would it be possible for ``evaluate`` and ``evaluate_batched`` to stick the arguments  into a tuple automagically?

ml_genn/model.py
jamesturner246: I've refactored `evaluate` to batch data and send off to `evaluate_batched`. Should be a bit cleaner now.

neworderofjamie (approved):
NICE! I think this is a big improvement

PR #44, opened 2022-04-08
New mlGeNN syntax

neworderofjamie:
I have been working away on this on the back burner for quite a while now and I think it has now reached the point where it can do everything the previous mlGeNN could do (and more) so I think it would make sense to get this stage reviewed. Up to @jamesturner246 whether we actually want to merge this into master or not I think.

# Syntax
## Model description
The syntax I described previously has evolved somewhat but a minimal example, defining a model in mlGeNN looks like:
```python
import numpy as np

from ml_genn import InputLayer, Layer, SequentialNetwork
from ml_genn.compilers import InferenceCompiler
from ml_genn.neurons import IntegrateFire, IntegrateFireInput
from ml_genn.connectivity import Dense

BATCH_SIZE = 128

# Create sequential model
network = SequentialNetwork()
with network:
    input = InputLayer(IntegrateFireInput(v_thresh=5.0), 784)
    Layer(Dense(weight=np.load("weights_0_1.npy")), 
          IntegrateFire(v_thresh=5.0))
    output = Layer(Dense(weight=np.load("weights_1_2.npy")), 
                   IntegrateFire(v_thresh=5.0, output="spike_count"))

compiler = InferenceCompiler(dt=1.0, batch_size=BATCH_SIZE, 
                             evaluate_timesteps=100)
compiled_net = compiler.compile(network)

# Load testing data
testing_images = np.load("testing_images.npy")
testing_labels = np.load("testing_labels.npy")

with compiled_net:
    # Evaluate model on numpy dataset
    metrics, cb_data  = compiled_net.evaluate({input: testing_images * 0.01}, {output: testing_labels})
    print(f"Accuracy = {100 * metrics [output].result}%")
```
The evaluate methods are a little smarter than the old versions and support ``Metric`` objects so you can calculate different metrics (currently "sparse_categorical_accuracy" and "mean_square_error" are supported).

## Callbacks and recording
In order to run custom logic mid-simulation, I've implemented a callback system (very similar to https://keras.io/api/callbacks/). Currently the ``evaluate_XXX`` methods of compiled models all take a list of callback objects (or the names of default-constructable callbacks in the same style as neuron models etc) which defaults to a list containing a ``callbacks.BatchProgressBar`` to show inference progress. However, you could instead add a ``VarRecorder`` callback to the above model:

```python
from ml_genn.callbacks import VarRecorder
...
callbacks = ["batch_progress_bar", VarRecorder(input, "V", key="v_input")]
metrics, cb_data = compiled_net.evaluate({input: testing_images * 0.01}, {output: testing_labels},
                                          callbacks=callbacks)
```
to record the input layer's V state variable over time. After the simulation has completed, you could then plot all 784 neurons membrane voltages in response to the first example with:
```python
import matplotlib.pyplot as plt
...
plt.plot(cb_data["v_input"][0])
plt.show()
```
Keys are used to uniquely identify recorded data produced by callbacks and can be any hashable type. If no key is provided, the integer index of the callback will be used e.g. in this case, the key of the ``VarRecorder`` would be ``1``.
Spike recording is very similar, you just need to use the ``callbacks.SpikeRecorder`` callback instead. The only caveat is that, as this uses GeNN's spike recording system (https://github.com/genn-team/genn/pull/372), you need to set the ``record_spikes=True`` keyword argument on ``Population``, ``InputLayer`` or ``Layer`` objects when you construct the model. For example:
```python
input = InputLayer(IntegrateFireInput(v_thresh=5.0), 784, record_spikes=True)
```

### Filtering
When you're dealing with large models/datasets, recording *everything* uses a lot of memory and slows the simulation down significantly. You can address this by adding _filtering_ kwargs to ``SpikeRecorder`` and ``VarRecorder`` objects. Example filters let you select which examples to record from:
```python
SpikeRecorder(input, example_filter=1000)    # Only record from example 1000
SpikeRecorder(input, example_filter=[1000, 1002]) # Only record from examples 1000 and 1002
SpikeRecorder(input, example_filter=[True]*10) # Only record from the first 10 examples
```

Similarly, neuron filters let you select which neurons to record from:
```python
SpikeRecorder(input, neuron_filter=1000)    # Only record from neuron 1000 in a 1D population
SpikeRecorder(input, neuron_filter=[1000, 1002]) # Only record from neurons 1000 and 1002 in a 1D population
SpikeRecorder(input, neuron_filter=[True]*10) # Only record from the first 10 neurons in a 1D population
SpikeRecorder(input, neuron_filter=np.s_[0::2]) # Only record from every other neuron in a 1D population
```
Because, in for example Convolutional neural networks, populations can have multidimensional shapes this syntax also extends to multiple dimensions in the same way as numpy arrays, for example:
```python
SpikeRecorder(input, neuron_filter=([16, 20], [16, 20])        # Record neurons(16,16) and (20, 20) in 2D population
SpikeRecorder(input, neuron_filter=np.index_exp[2:4,2:4])   # Record neurons (2,2), (2,3), (3,2) and (3,3) in 2D population
```

## Metrics
mlGeNN can calculate different types of metric when evaluating datasets. In the above example, the default ``SparseCategoricalAccuracy`` metric is calculated for all output neurons but this can be overridden. For example, in a regression task you could instead do:
```python
metrics = compiled_net.evaluate({input: x}, {output: y}, "mean_square_error")
```
to calculate a mean square error. Finally, if a model has multiple outputs, different metrics can be calculated for each one by providing a dictionary of metrics:
```python
metrics = compiled_net.evaluate({input: x}, {output: y, output_2: y2}, {output: "mean_square_error", output_2: "sparse_categorical_accuracy"})
print(f"Error = {metrics[output].result}")
print(f"Accuracy = {100 * metrics[output_2].result}%")
``` 

# TensorFlow
The TensorFlow dependency is split out into the new ``ml_genn_tf`` module which supports the existing conversion functionality with syntax something like this:

```python
# Create suitable compiler for model
converter = FewSpike(k=10, signed_input=True, 
                     norm_data=[norm_data])

# Convert and compile ML GeNN model
net, net_inputs, net_outputs, tf_layer_pops = converter.convert(tf_model)

compiler = converter.create_compiler()
compiled_net = compiler.compile(net, inputs=net_inputs, outputs=net_outputs)
```
The nice thing here is that you can modify the converted model after conversion step, using the ``tf_layer_pops `` dictionary to get the mlGeNN populations corresponding to each layer, and modify the GeNN model after compilation.

I'm not sure if reviewing this by GitHub PR review is really practical but, if you do have time to look through it, that would be wonderful and I would be more than happy to talk you both through it when @tnowotny is back from holiday/incorporate any feedback you might have. Also, @jamesturner246 , if you could check your models convert properly and perform the same, that would be wonderful - the CIFAR10 examples have all been updated so you can see how it interfaces with TF datasets.

TODO
- [ ] Update ImageNet models
- [x] Automatic unique model naming
- [x] Keras-style callback system e.g. for recording also maybe progress bars


tnowotny:
Yeah, I am not sure whether reviewing this straight on github as a PR will be meaningful. But in the meantime two comments:
1. I see the data from the callbacks is indexed by integers in your example; maybe it needs to be that way for generality but for data recorders indexing by a variable name or such might have been more intuitive.
2. I think in the filter section the last two examples have problems:
a) did you mean to use neuron_filter?
b) I am not sure I understand ```np.index_exp[2:4,2:4]``` and your expansion of it ... would it not be functionally the same as ```[2,3], [2,3]``` and hence record neurons (2,2) and (3,3)?

neworderofjamie:
> Yeah, I am not sure whether reviewing this straight on github as a PR will be meaningful. But in the meantime two comments:
> 
> 1. I see the data from the callbacks is indexed by integers in your example; maybe it needs to be that way for generality but for data recorders indexing by a variable name or such might have been more intuitive.

So the index is into the array of callbacks as each recorder only records one variable but, nonetheless, I'm not totally happy about this bit of API either. I think for consistency, it t should work more like the metrics where the results get returned in some sort of dictionary 
> 2. I think in the filter section the last two examples have problems:
>    a) did you mean to use neuron_filter?

Where do you mean? 

>    b) I am not sure I understand `np.index_exp[2:4,2:4]` and your expansion of it ... would it not be functionally the same as `[2,3], [2,3]` and hence record neurons (2,2) and (3,3)?

Ahh yes, that bit of numpy magic was a new discovery for me too. Basically, it's shorthand for ``(slice(2,4), slice(2,4)`` so it represents the intersection of those two per-dimension slices

tnowotny:
I think in the filter section the last two examples have problems:
a) did you mean to use neuron_filter?
Where do you mean?
```SpikeRecorder(input, example_filter=([16, 20], [16, 20])        # Record neurons(16,16) and (20, 20) in 2D population
```
if it's an example_filter, why would it index neurons?

The ```index_exp``` magic is still confusing - so a pair of lists is combined entry by entry as in the example just above: ``` [16, 20], [16, 20]``` -> (16,16),(20,20)
but the tuple of slices ```np.index_exp[2:4,2:4] == (slice(2, 3, None), slice(2, 3, None))``` is combined all-to-all?

neworderofjamie:
> I think in the filter section the last two examples have problems: a) did you mean to use neuron_filter? Where do you mean?
> 
> ```
> SpikeRecorder(input, example_filter=([16, 20], [16, 20])        # Record neurons(16,16) and (20, 20) in 2D population
> ```
> 
> if it's an example_filter, why would it index neurons?

Oh yeah, I see it now - fixed 

> 
> The `index_exp` magic is still confusing - so a pair of lists is combined entry by entry as in the example just above: ` [16, 20], [16, 20]` -> (16,16),(20,20) but the tuple of slices `np.index_exp[2:4,2:4] == (slice(2, 3, None), slice(2, 3, None))` is combined all-to-all?

Yep, it should work the same as it does it in numpy

tnowotny:
... I am probably just a bit behind the curve, but:
```python
import numpy as np
x= np.index_exp[2:4,2:4]
y= np.array([[1,2,3,4],[5,6,7,8],[9,10,11,12],[13,14,15,15]])

print(x)
print(y)
print(y[x])
```
```
(slice(2, 4, None), slice(2, 4, None))

[[ 1  2  3  4]
 [ 5  6  7  8]
 [ 9 10 11 12]
 [13 14 15 16]]

[[11 12]
 [15 16]]```

neworderofjamie:
> ok, never mind ... though on reflection I find the example with the pair of lists more confusing - are you sure you want [row1, row2], [col1, col2] instead of [row1, col1], [row2,col2]?

Again, personally I would agree, but numpy indexing of all sorts (aside from when you're using a mask) works that way

PR #53, opened 2023-02-28
When using VarRecorder you shouldn't have to know the internal GeNN variable names

### LINKED ISSUE #47: VarRecorder should use friendly python variable names rather GeNN ones
This should be trivial with ``ValueDescriptor`` stuff

neworderofjamie:
In mlGeNN model descriptions, a ``ValueDescriptor`` decorator is used to handle a lot of variable initialisation boilerplate (basically they are like supercharged properties that define what happens when you get and set the attribute). It also manages the mapping of python names to the names used in GeNN e.g. in the following:
```python
class IntegrateFire(Neuron):
    v = ValueDescriptor("V")
    ...
```
``v`` will get instantiated as a GeNN variable called ``V``. Previously you had to figure out the name of the GeNN variable when using a ``VarRecorder`` but this PR adds a helper that can do the looking up and uses that. Pretty sure this is the only outward-facing API change on the list of known issues so think I will make a release after this. In order to still allow recording of 'internal' state that might have been added by compilers, you can still use the ``genn_var`` keyword argument to the ``VarRecorder`` constructor.

Fixes #47


ml_genn/ml_genn/callbacks/var_recorder.py
```diff
@@ -2,22 +2,29 @@
...
         # Get underlying population
         # **TODO** handle Connection variables as well
         self._pop = get_underlying_pop(pop)
-        self._var = var
+
+        # Get the name of the GeNN variable corresponding to var
+        if var is not None:
+            self._var = get_genn_var_name(self._pop.neuron, var)
+        elif genn_var is not None:
+            self._var = genn_var
```
tnowotny, line 27: what happens if neither var nor genn_var are not None?

tnowotny (commented):
in general yes, all good - but in original version and new version what happens if no variable is specified?

ml_genn/ml_genn/callbacks/var_recorder.py
neworderofjamie: Excellent point, will add an error and then merge

tnowotny: ok, approved then :-)

PR #54, opened 2023-03-13
Spike event recording

neworderofjamie:
It would have been marginally nicer to have a derived callback class for spike-event recording but that would have broken the [slightly primitive mechanism](https://github.com/genn-team/ml_genn/blob/master/ml_genn/ml_genn/utils/callback_list.py#L36-L38) for calling ``set_first`` on one callback of each type (used to only pull recording buffers once)


tnowotny (approved):
Looks indeed straightforward.

PR #63, opened 2023-04-28
Compiler defaults

### LINKED ISSUE #59: Smarter defaults
Currently, when you construct a population of neurons e.g.
```python
LeakyIntegrateFire(v_thresh=1.0, tau_mem=20.0, tau_refrac=None, relative_reset=False, integrate_during_refrac=False, scale_i=True)
```
as well as settings a user might well want to set like ``v_thresh`` and ``tau_mem``, the ones like ``integrate_during_refrac`` are pretty annoying. They definitely belong as part of the network description as they need to be consistent between training and inference compilers but it's very annoying having to remember which ones are right to compile your network for a given compiler.  Compilers could provide a dictionary of defaults which could be passed to the network constructor and  _somehow_ get applied in the neuron class constructor.

neworderofjamie:
One really annoying thing with this world is that e.g. LIF neurons are not defined in a uniform way between mathematical frameworks. For example, e-prop uses relative reset and doesn't integrate during the refractory period whereas EventProp does neither of these things. Not only does this mean users had to set a bunch of random flags when defining a model (see ``relative_reset`` and ``integrate_during_reset`` in the following) but they also had to change them if they switched from e.g. e-prop to EventProp.

```python
network = SequentialNetwork()
    LeakyIntegrateFire(v_thresh=0.61, tau_mem=20.0,
                                        tau_refrac=5.0,
                                       relative_reset=True,
                                       integrate_during_refrac=True,
                                       tau_refrac=5.0),
```
Now each compiler can provide a dictionary of neuron/synapse/whatever types to a dictionary of default parameters for their constructor parameters so the user can do this instead:
```python
from ml_genn.compilers.eprop_compiler import default_params

network = SequentialNetwork(default_params)
    LeakyIntegrateFire(v_thresh=0.61, tau_mem=20.0,
                                        tau_refrac=5.0,
                                       tau_refrac=5.0),
```

Internally, it is a little gnarly but the default parameters get stashed in the ``Network``/``SequentialNetwork`` object and pulled out in the ``network_default_params`` decorator which is applied to model constructor. This wraps itself around the constructor and adds these 'runtime defaults' to the specified parameters.

Fixes #59

### DIFF
--- a/examples/eprop/latency_mnist.py
+++ b/examples/eprop/latency_mnist.py
@@ -14,6 +14,8 @@ from time import perf_counter
 from ml_genn.utils.data import (calc_latest_spike_time, calc_max_spikes,
                                 log_latency_encode_data)
 
+from ml_genn.compilers.eprop_compiler import default_params
+
 NUM_INPUT = 784
 NUM_HIDDEN = 128
 NUM_OUTPUT = 10
@@ -29,7 +31,7 @@ spikes = log_latency_encode_data(
     20.0, 51)
 
 serialiser = Numpy("latency_mnist_checkpoints")
-network = SequentialNetwork()
+network = SequentialNetwork(default_params)
 with network:
     # Populations
     input = InputLayer(SpikeInput(max_spikes=BATCH_SIZE * calc_max_spikes(spikes)),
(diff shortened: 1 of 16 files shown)


tnowotny (approved):
Yeah, this looks sensible though I have nit fully analysed the way it's implemented with the wrapping.

PR #64, opened 2023-04-28
Per-timestep EventProp loss

neworderofjamie:
As promised, this PR re-adds per-timestep loss to EventProp. Because it adds a chain of softmax custom updates every timestep, it is significantly slower than the per-batch loss functions.

Note, this requires https://github.com/genn-team/genn/tree/origin/egp_ref which I will also make a PR for.

### DIFF
--- a/ml_genn/ml_genn/compilers/compiler.py
+++ b/ml_genn/ml_genn/compilers/compiler.py
@@ -198,7 +198,8 @@ class Compiler:
                           group: str, name: str):
         # Process model
         (cu_model, cu_param_vals, cu_var_vals,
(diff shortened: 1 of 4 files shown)


tnowotny (approved):
Loks good overall though a difficult one to parse. I did not check details fo indexing into the ring buffers but would trust that if an example (MNIST?) appears to run fine (expect good accuracy on MNIST).

neworderofjamie:
Just verified everything is working on my machine and obtained the following after 10 epochs on my machine:
||Train accuracy [%]|Test accuracy [%]|Train time[s]|
|-|-|-|-|
|Per-timestep|96.35|96.6|50
|Per-trial|96.5|96.4|28

PR #65, opened 2023-04-28
Bugfix for strange callback behaviour with validation 

neworderofjamie:
So this was caused by callbacks (in this case spike recorders) only getting shallow-copied into the ``CallbackList`` objects used internally to dispatch callbacks. This meant two spike recorders were adding data to the same data structure resulting in what you were seeing. The solution to this was simply to make ``CallbackList`` own the state itself  which was easy to do as it already was responsible for delivering callback data back to the ``train`` method.


tnowotny (approved):
Ok, this one's straightforward - looks good.

PR #66, opened 2023-04-28
Compiled network fixes

neworderofjamie:
This fixes two annoying things:
* Using https://github.com/genn-team/genn/pull/581 it unloads models when the ``CompiledNetwork`` context manager is left
* When you enter the ``CompiledNetwork`` context manager, it only builds the model if it's not already built (maybe this should be PyGeNN's default behaviour tbh)


tnowotny (approved):
Aha, not that difficult to fix actually - looks good to me.

neworderofjamie:
Great! Will merge once https://github.com/genn-team/genn/pull/581 is also merged (I feel master of mlGeNN should at least work with master of GeNN)

PR #69, opened 2023-07-17
Time varying inputs

neworderofjamie:
Previously mlGeNN support raw spiking inputs and various ways to turn a single 'frame' into one example-worth of spiking activity. However, sometimes you want to turn multiple frames into spiking activity within a single example e.g. Mel spectograms. 

This PR refactors most neuron inputs to support this. Main change is taht ``InputBase`` now has additional functionality to add either a state variable (for non-time varying input) or an extra global parameter (for time-varying input) to a neuron model and generate correct indexing code. This also lets standard neuron model e.g. ``IntegrateFire`` and ``LeakyIntegrateFire`` be re-used as input models trivially.

In order to use this functionality to e.g. present groups of 10 MNIST digits to an input layer of ``IntegrateFireInput`` neurons at 100 timestep intervals, you would configure IntegrateFireInput like this:
```python
IntegrateFireInput(v_thresh=5.0, input_frames=10, input_frame_time=100)
```
Then you would reshape the images appropriately like this:
```python
testing_images = np.reshape(mnist.test_images(), (-1, 10, 784))
```
and, finally, after setting the ``evaluate_timesteps`` to 1000 you would see spiking output like this:
Figure_2

### DIFF
--- a/ml_genn/ml_genn/compilers/compiler.py
+++ b/ml_genn/ml_genn/compilers/compiler.py
@@ -177,7 +177,8 @@ class Compiler:
 
         # If source neuron model defines a negative threshold condition
         src_pop = connection.source()
-        src_neuron_model = src_pop.neuron.get_model(src_pop, self.dt)
+        src_neuron_model = src_pop.neuron.get_model(src_pop, self.dt,
+                                                    self.batch_size)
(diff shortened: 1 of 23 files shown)


tnowotny (approved):
I like it. Syntax seems fine and I could follow roughly how it works. Looks good I think.

PR #70, opened 2023-07-17
Exact update for output neuron lambda_v and lambda_i

neworderofjamie:
Implemented changes in https://github.com/tnowotny/genn_eventprop/commit/212521e175489503fada4b9eebbe2e5a22b527aa in mlGeNN EventProp compiler. SHD and latency MNIST examples definitely still learn (using ``AvgVarExpWeight`` and ``AvgVar`` output respectively). 

Code should be easy to compare to original - only thing to note is that because we're _prepending_ a backward pass to the beginning of the neuron model, the order of code generation is backwards.

### DIFF
--- a/ml_genn/ml_genn/compilers/event_prop_compiler.py
+++ b/ml_genn/ml_genn/compilers/event_prop_compiler.py
@@ -401,12 +401,19 @@ class EventPropCompiler(Compiler):
                 tau_mem = pop.neuron.tau_mem
                 model_copy.add_param("TauM", "scalar", tau_mem)
                 model_copy.add_param("A", "scalar", 
-                                     tau_mem / (tau_syn - tau_mem))
+                                     tau_mem / (tau_mem - tau_syn))
 
                 # Add extra global parameter to contain trial index and add 
                 # population to list of those which require it updating
                 model_copy.add_egp("Trial", "unsigned int", 0)
                 compile_state.update_trial_pops.append(pop)
+
+                # Prepend standard code to update LambdaV and LambdaI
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
Yes, I think this is accurate. But looking through it, it occurs to me that this does not explain why the mlGeNN version was learning well enough with fewer timesteps/ large DT and the genn_eventprop version wasn't. Both of them appear to have used the same simple Euler step instead of the now more accurate exact solution. Bit mysterious. I saw your comment about learning slower; adjusting the learning rate could be an answer. But overall I am still a bit weary about the discrepancies.

neworderofjamie:
Recombining our respective memories, I don't think this was quite true. The mlGeNN version previously used a _broken_ Exponential Euler whereas the PyGeNN version used a _correct_ Forward Euler so I think the mystery is solved.

tnowotny:
Oh yes, to complete the record of what we found, the issue was that lambda_V-lambda_I (which determines the amplitude of the gradient flowing from output to hidden layer) converges to 0 as a function of time -> -infinity for the constant loss term when using the "sum" loss function. Numerically, in the different versions, the explicit linear Euler (or "forward" linear Euler) algorithm used in genn_eventprop converged fairly close to 0 (though also plateauing on a small value above 0 as I recall it) but the "botched" exponential Euler in mlGeNN plateaued at a value significantly larger than 0 which allowed more gradient flow towards the hidden layer. This explained the vastly different scales of gradients in the hidden layer (and related vastly different sizes of regularization terms needed).

PR #72, opened 2023-08-18
Fixed recording of VarAccess::SHARED_NEURON variables

neworderofjamie:
This functionality is clearly newer than my implementing of ``VarRecorder``. Nonetheless, this seems to work


tnowotny (approved):
Looks ok - if it works! So I'll approve.

PR #73, opened 2023-08-31
Spike count record

### LINKED ISSUE #60: Spike count recorder
Often spike counts are all you actually want to record - probably easiest to make this a flag for ``SpikeRecorder``  as implementing this would only require a small change to ``on_timestep_end`` i.e. ``np.sum(data_unpack[:, b, self._neuron_mask])`` rather than the current ``np.where``.

neworderofjamie:
Implemented spike count recording as an option for the ``SpikeRecorder`` callback meaning that all existing filtering etc works here. The implementation could use https://github.com/numpy/numpy/pull/21429 to further improve efficiency.

```python
# Evaluate model and record spikes and spike counts of input population
callbacks = ["batch_progress_bar", SpikeRecorder(input, key="input_spikes"),
             SpikeRecorder(input, key="input_spike_counts", record_counts=True)]
metrics, cb_data  = compiled_net.evaluate({input: spikes},
                                          {output: labels},
                                          callbacks=callbacks)

# Extract spikes and spike counts
input_spikes = cb_data["input_spikes"]
input_spike_counts = cb_data["input_spike_counts"]

# Assert that manually-calculated spike counts from arbitrary example match those calculated using new system
assert np.array_equal(np.bincount(input_spikes[1][100], minlength=NUM_INPUT),
                                  input_spike_counts[100])

```
Fixes #60

### DIFF
--- a/examples/eprop/latency_mnist.py
+++ b/examples/eprop/latency_mnist.py
@@ -22,9 +22,10 @@ NUM_OUTPUT = 10
 BATCH_SIZE = 128
 NUM_EPOCHS = 10
 SPARSITY = 1.0
-TRAIN = True
+TRAIN = False
 KERNEL_PROFILING = False
 
+np.seterr(all='raise')
 labels = mnist.train_labels() if TRAIN else mnist.test_labels()
 spikes = log_latency_encode_data(
     mnist.train_images() if TRAIN else mnist.test_images(),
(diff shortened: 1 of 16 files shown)


tnowotny (approved):
Ok, uncontroversial this one ;-)

PR #74, opened 2023-09-04
Non-change to enable augmentation

### LINKED ISSUE #58: Data augmentation
Not 100% sure where this should be integrated:
- Some things e.g. random shift or time dilation _could_ be added to ``SpikeSourceInput``
- It would be nice to support tonic's existing augmentation library

neworderofjamie:
As described in #58, we don't need to add anything to enable augmentation while training. However, this PR fixes a small bug which meant the loop over epochs in ``CompiledTrainingNetwork`` was incorrect and adds some examples of augmented training workflows.

Fixes #58

### DIFF
--- a/examples/event_prop/latency_mnist_augment.py
+++ b/examples/event_prop/latency_mnist_augment.py
@@ -1,0 +1,118 @@
+import numpy as np
+import mnist
+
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
Ok, this one is the truly unproblematic one!

PR #83, opened 2023-12-01
Fix width height confusion in tonic sensor_size 

tnowotny:
Assuming that the tonic shape is width, height, channels, y needs to be multiplied with shape[0] (the width).


neworderofjamie:
I can verify, based on https://github.com/neuromorphs/tonic/blob/develop/tonic/datasets/asl_dvs.py#L39 and https://github.com/neuromorphs/tonic/blob/develop/tonic/datasets/visual_place_recognition.py#L69 that ``shape[0]`` is indeed width

neworderofjamie (approved):

PR #85, opened 2023-12-07
warn if no key is provided for a recorder

tnowotny:
just using a default integer key is very confusing.


neworderofjamie (commented):
I think you need a ``from warnings import warn`` at the top

tnowotny:
Yes ... good point. I probably should have done a minimal test whether the proposed change actually work ...
Surprisingly all tests passed - so this case is not covered ... import now added

neworderofjamie (approved):

PR #86, opened 2024-01-26
Spike time quantisation

neworderofjamie:
Spike data coming from tonic often can easily have multiple events assigned to one neuron in a timestep, especially if the ``Downsample`` transformation is applied. Previously, ``preprocess_tonic_spikes`` converted the tonic events directly to spike-source array format but this PR adds an option to apply a 2D (neuron id, timestep) histogram and threshold this to obtain events. ``histogram_thresh = 1`` will simply remove duplicate spikes and higher numbers will counter-act the noise amplification effect of the ``Downsample``

Also, as I didn't want to make another PR, this includes a fix for a bug in ``LeakyIntergrator`` output neuron model where bias term was never being applied - this really degrades e-prop performance.

### DIFF
--- a/ml_genn/ml_genn/neurons/leaky_integrate.py
+++ b/ml_genn/ml_genn/neurons/leaky_integrate.py
@@ -31,8 +31,8 @@ class LeakyIntegrate(Neuron):
 
         # Define integration code based on whether I should be scaled
         if self.scale_i:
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
Yes, totally agree, this spike duplication issue is bad and this is an easy solution for users to use.

PR #93, opened 2024-04-19
GeNN 5

neworderofjamie:
I'm afraid this is a bit of a mess. It's a combination of lots of minor syntax updates to make mlGeNN compatible with GeNN 5 and loads of documentation. Probably most meaningful to look at the documentation https://ml-genn.readthedocs.io/en/genn_5/


tnowotny (changes requested):
in "building_networks": TODO link Many other aspect

in "even_prop compiler": Is rge_nu_upper indeed in Hz?

Is this still true: If no key is provided, the integer index of the callback will be used e.g. in this case, the key of the SpikeRecorder would be 1?

in "callback_recording": Incomplete:
"To give them access to properties of the 
data
compiled_network
num_batches
num_epochs" 

I found "called at the start of training on epoch ``epoch``" slightly confusing. It's called at the start of every epoch, passing the epoch number in ``epoch``, right? Maybe there is a better way of saying this.

"mean_squared_error" would have been more appropriate than "mean_square_error" (too late to change that?)

 "converting_tf": Incomplete/ ends on: "For example, to enable spike recording on a"

Finally, why is the copyright "James Turner, Jamie Knight" rather than the other way round?

PS: I also fixed some typos that I will push directly into the genn_5 branch - so pull first!

docs/usage/building_networks.rst
```diff
@@ -47,9 +47,8 @@ By default, :class:`~connection.Connection` objects use a 'delta' synapse model
 accumulated weight of incoming spikes is directly injected into neurons. However, if
 you wish to use a somewhat more realistic model where inputs are *shaped* to mimic the 
 dynamics of real ion channels, this can be replaced. 
-This same general principle is also used for configuring many other aspects of mlGeNN,
-including loss functions, metrics and readouts. TODO link
-Many other aspec
+This same general principle is also used for specifying and configuring many other aspects of mlGeNN networks,
+including :mod:`ml_genn.losses`, :mod:`ml_genn.metrics and :mod:`ml_genn.readouts`.
```
tnowotny, line 51: is there a tick missing after metrics?

tnowotny (approved):
I think I caught a last typo. Otherwise it seems good!

PR #94, opened 2024-04-19
Deep R

neworderofjamie:
This is the long-deferred (see the number of force push rebases below 😱) DEEP R implementation for mlGeNN - I'd really like to merge it so we can make a final(?) 2.4 release to alongside the paper (the additional loss function for Balazs' paper is also currently unreleased). DEEP R works reasonably well with e-prop but, with Eventprop it doesn't for reasons I have long struggled to understand. What do you think is the best solution? I could fully unhook DEEP R from the Eventprop compiler so it doesn't even look like it's possible to use or add a warning saying it doesn't work or just leave it.

This PR also also makes the existing ``CustomUpdate`` callback more flexible so you don't need to define custom callback classes every time you e.g. want a callback which triggers on the last but one timestep of a batch or whatever.


tnowotny (approved):
It looks sensible and I don't see an obvious reason why it wouldn't work with eventprop equally well. For the purposes of moving forward, I will approve this PR and we keep an open issue to look into why Eventprop + deep R doesn't seem to work.

PR #96, opened 2024-06-18
Several bug fixes and additional tests

neworderofjamie:
Clearly I was in a bit of a rush releasing the new version of mlGeNN corresponding with GeNN 5 before Capocaccia because Balazs found several stupid bugs in spike recording. This PR fixes those and a slightly subtler one that broke regularisation with EventProp due to lambda values being cast to integer. I have also added a test of spike recording functionality.


tnowotny (approved):
All makes sense.

PR #103, opened 2024-07-01
fix egp index for input current with  batch size 1

FabianSchubert:
For batch size 1, `t` instead of `timestep` was used to read from the egp, which caused an error in the compilation of the GeNN model. It should use `timestep` as in the case where `batch_size > 1`.


neworderofjamie (approved):
Good spot

PR #104, opened 2024-07-02
Delays and EventProp

### LINKED ISSUE #90: user callbacks with online validation
This currently behaves strangely with e.g. ``on_epoch_begin`` callback being called once for the training epoch and once for the subsequent validation 'epoch.

When training, Keras does:

1. on_train_begin
2. on_epoch_begin
3. on_train_batch_begin
4. on_train_batch_end
5. on_test_begin
6. on_test_batch_begin
7. on_test_batch_end
8. on_test_end
9. on_epoch_end
10. on_train_end

5-8 only happen when doing validation.
When testing, only 5-8 occur.

neworderofjamie:
This PR improves handling of delays and allows them to be learned using EventProp (using new functionality from https://github.com/genn-team/genn/pull/629).

However, incorporating delays means that you need to be careful between trials to empty dendritic delay buffers (as well as ``outPost``) and make sure variables accessed with delays (``LambdaI``) are also reset correctly. This PR modifies the existing reset logic using the new functionality implemented in https://github.com/genn-team/genn/pull/634, namely:

* The ``create_reset_custom_update`` helper now uses ``VarAccessMode.BROADCAST`` wherever possible so resets will be applied across all delay slots if necessary (also saves an unnecessary memory read)
* Added new ``Compiler.add_out_post_zero_custom_update`` helper to create custom updates that zero ``outPost`` and, if necessary ``denDelay`` on device.

This PR also includes fixes for several issues which somehow only came to light when trying to train SSC.

* Gradients were being optimised and reduced in 1st batch when they are invalid
* During validation, gradients are still accumulated - this is fine but they weren't being zeroed before the next epoch of training. An additional custom update is now generated to do this.

Also, this PR modifies ``CompiledTrainingNetwork.train`` to take a seperate list of validation callbacks, preventing irritating issues around e.g. checkpointing and learning rate schedules which I think Tom, Balazs and everyone else who's tried to use mlGeNN hit!

Fixes #90


tnowotny (approved):
I think I was able to follow most of it and it looks ok. Just to confirm, delays are *not* reset after an epoch, right? Only the gradients either for weights or for delays. (otherwise the reset would be fatal if learning delays, obviously).

PR #115, opened 2024-09-03
EventProp fixes

neworderofjamie:
This branch contains several EventProp fixes that have emerged from my recent debugging deep dive:
* LIF + ExpCurr neuron model were not matching Thomas's code with exact integration. Slightly extended ``ValueDescriptor`` system to enable multiple derived GeNN variables to be calculated from a single Python parameter to tidily implement scale factor already present in GeNN ExpCurr model. This is configured via the standard compiler ``default_params`` mechanism.
* I think because we reset time every batch, the ring buffer is not empty at the start of each epoch like it is in Thomas's code. To address this we move the new gradient zeroing update to the end of batch 1 to ensure gradients are accumulated correctly.
* Rather than using the per-neuron spike counts for regularisation, Thomas's code reduced these over all batches. This functionality has now been implemented,

### DIFF
--- a/examples/event_prop/shd.py
+++ b/examples/event_prop/shd.py
@@ -21,8 +21,7 @@ from ml_genn.compilers.event_prop_compiler import default_params
 
 NUM_HIDDEN = 256
 BATCH_SIZE = 32
-NUM_EPOCHS = 300
-EXAMPLE_TIME = 20.0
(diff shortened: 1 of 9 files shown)


tnowotny (approved):
That looks sensible. Great work in pinning down those subtle "bugs".

PR #117, opened 2024-09-17
Yin yang dataset, TTFS readout and event-prop implementation

### LINKED ISSUE #107: Time to first spike
This would be a new readout and EventProp loss function implementation (some sort of EventProp implementation at https://github.com/tnowotny/genn_eventprop/blob/main/models.py#L1265-L1338)

neworderofjamie:
* Added yin yang dataset generation based on https://github.com/lkriener/yin_yang_data_set as it's not packaged anywhere so annoying to include
*  Added ``FirstSpikeTime`` readout which produces time of first spike for each neuron. Happy to debate this design decision but, this way, standard metrics and readouts can be applied.
* Added support for LIF output neurons with ``FirstSpikeTime`` readout to EventProp compiler
* Added EventProp example training on YinYang based on @tnowotny's https://github.com/tnowotny/genn_eventprop/blob/main/train_yinyang.py

Fixes #107

### DIFF
--- a/ml_genn/ml_genn/compilers/event_prop_compiler.py
+++ b/ml_genn/ml_genn/compilers/event_prop_compiler.py
@@ -850,9 +850,6 @@ class EventPropCompiler(Compiler):
                 
                 # Add variable to hold backspike flag
                 model_copy.add_var("BackSpike", "uint8_t", False)
-                
-                # Add variable to hold phantom backspike flag
-                model_copy.add_var("BackPhantomSpike", "uint8_t", False)
 
                 # Add EGP for spike time ring variables
                 ring_size = self.batch_size * np.prod(pop.shape) * self.max_spikes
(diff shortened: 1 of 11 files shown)


ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -813,6 +820,138 @@ def build_neuron_model(self, pop: Population, model: NeuronModel,
...
+                        """
+
+                    # On backward pass transition, update LambdaV if this is the first spike
+                    # **THOMAS** why are we dividing by what looks like softmax temperature?
+                    transition_code = f"""
+                        const scalar iMinusVRecip = 1.0 / RingIMinusV[ringOffset + RingReadOffset];
+                        LambdaV += iMinusVRecip * (Vthresh * LambdaV);
+                        if (fabs(backT + TFirstSpikeBack) < 1e-3*dt) {{
+                            if (id == YTrueBack) {{
+                                const scalar fst = {example_time} - TFirstSpikeBack;
```
tnowotny, line 907: I am not sure about this formula. I think it might have to be just "-TFirstSpikeBack"

tnowotny (commented):
Is this working well? I think there might be an error in the "fst" formula as noted in my comment.

ml_genn/ml_genn/compilers/event_prop_compiler.py
neworderofjamie: so in your code it was:
```
scalar fst= $(first_spike_t)-$(rev_t)+$(trial_t);
```
However, because in mlGeNN, trials always start at time 0, ``rev_t=0``  transforming to:
```
scalar fst= $(first_spike_t)+$(trial_t);
```
and first spike times are negated which transforms to:
```
scalar fst= $(trial_t) - $(first_spike_t);
```

tnowotny: ```rev_t == trial_t``` in the ml-genn scheme, hence ```fst = first_spike_t - trial_t +trial_t``` 
then negated, it becomes ```fst = -TFirstSpikeBack```

PR #125, opened 2024-12-05
Increased max delay

### LINKED ISSUE #113: Longer static delays
Should use logic like GeNN does for sparse connectivity indices and switch to ``uint16_t`` if max is more than 255.

neworderofjamie:
Heterogeneous delays were previously always stored in ``uint8_t``, limiting them to a maximum of 255 timesteps. @mbalazs98 code this is based on _always_ used ``uint16_t`` which is wasteful.

This PR implements automatic switching to ``uint16_t`` if maximum delay is >= 256

Fixes #113


tnowotny (approved):
I think it's good.

PR #131, opened 2025-02-10
Numerous typos in loading delays

neworderofjamie:
Loose ends from #125 but
* Max delay should always be an integer
* Typos


tnowotny (commented):
yes, makes sense ... meant to hit "approve".

tnowotny (approved):

PR #133, opened 2025-02-19
Auto-adjoint refactor

neworderofjamie:
Obviously many loose ends but I think this probably works as well as @tnowotny 's version and it's always good to create smaller PRs. 

Basic design is, as well as the helper classes for directly building and manipulating GeNN models (``NeuronModel``, ``SynapseModel``, ``WeightUpdateModel`` etc), I have added ``AutoNeuronModel`` and ``AutoSynapseModel`` which let you specify models in hybrid system language. Basic design is that ``Neuron.get_model`` and ``Synapse.get_model`` can return **either** and, as such, standard neuron and synapse now produce 'auto' models. I have also added ``UserNeuron`` and ``UserSynapse`` models for prototyping models in a similar style to the previous ``AutoSyn`` and ``AutoNeuron`` models - these are demonstrated in the ``latency_mnist_user.py`` example.

When given a GeNN model, ``Compiler.build_neuron_model`` and ``Compiler.build_synapse_model`` just create a copy as before but, when given an 'auto' model, they solve the equations and build a GeNN model containing the forward system. The Eventprop compiler builds on this, generates the adjoint system and adds it to the GeNN model in a similar way to before. 

There are a small number of ``**THOMAS**`` points for discussion either here on in person!

I have also started removing things from mlGeNN that are no longer necessary as the plan is for EventProp to be the only supported training algorithm for mlGeNN 3:

* eProp - computational cost of training these models is totally dominated by synapse dynamics so GeNN doesn't really have any advantage here so something like https://arxiv.org/abs/2501.11407 is probably a better idea.
* Removed default parameters system and applied Eventprop defaults the manual way - confused almost everyone and no longer necessary as auto-eventprop should handle e.g. different sorts of reset
* Refractory time from built in LIF-like models - I feel this is probably _possible_ to implement in Eventprop but I don't really think it's necessary
* Removed the separation between friendly and GeNN variable names - also confusing and who cares whether variable names in generated code are snake case.

TODO:
- [ ] Special case for delta synapses (implement as non-auto synapses and do the sensible thing)
- [ ] Delay learning
- [ ] Better regularisation or, as a minimum, scale current regularisation if standard LIF neuron is detected


tnowotny (commented):
I think I take this in turns, so here are a couple of comments after inspecting the resulting UserNeuron and UserSynapse use.
I was wondering
1. is ```output_var_name``` a property of a neuron or of the population/layer?
2. is there any role for ```output_var_name``` in neurons that are not output neurons?
3. Similarly, "readout" was for me more of a property of the population/layer as it relates to the network output handling, less to individual neurons. And is related to the loss, which I didn't think of as a neuron property. But I appreciate that practically it might not make a difference because our populations have homogeneous neuron type, so a neuron property can be interpreted as a layer property and vice versa.

neworderofjamie:
> I think I take this in turns, so here are a couple of comments after inspecting the resulting UserNeuron and UserSynapse use. I was wondering
> 
> 1. is `output_var_name` a property of a neuron or of the population/layer?
Neuron model - an ALIF neuron's output is always going to be it's V state variable not it's adaptation variable but it could be called anything

> 2. is there any role for `output_var_name` in neurons that are not output neurons?
As of my last change, yes - it's also used for output loss/regularisation injection

> 3. Similarly, "readout" was for me more of a property of the population/layer as it relates to the network output handling, less to individual neurons. And is related to the loss, which I didn't think of as a neuron property. But I appreciate that practically it might not make a difference because our populations have homogeneous neuron type, so a neuron property can be interpreted as a layer property and vice versa.
> It is implemented as a neuron property but, now you say it, I think it might make more sense as a ``Population`` thing. The link to loss is also a bit weird - in _theory_ it just provides the input to the loss function but in Eventprop the two are a bit inter-twined. This area of the design needs more thought.

tnowotny:
> is there any role for output_var_name in neurons that are not output neurons?
As of my last change, yes - it's also used where output loss/regularisation gets injected

Ah, yes. That is a good point. But thinking about it, our regularisation is based on spike numbers. Spikes are regulated through the threshold condition. So, ideally, for Eventprop there should be a regularisation term in the loss function that would depend on the vars that determine the threshold condition. And one could determine which var gets a reg term injected by dl/dx where l is the loss and x is the relevant var. Unfortunately, spike-number-based reg terms can't be translated into such a loss formalism, so our regularisation is ad-hoc. 
I am not offering solutions I realise but to identify an output var as the target for regularisation terms might not always be appropriate?

neworderofjamie:
Based on this, constraint wouldn't regularisation actually be injected into LambdaV and LambdaAdaptation for our hypothetical ALIF neuron? Further evidence we need to do more work on a more principled regulariser

tnowotny:
yes, I think injecting into both LambdaV and LambdaAdaptation (with opposite signs) would be correct. We wprobably should do some work on more principled regularisation but when I last tried I was unable to crack to maths of spike-number-based regularisation coming to the conclusion that it is actually impossible to do properly in the adjoint formalism. A possible way out would be to revisit voltage-based regularisation ideas (I *think* you mentioned you found something about it in Chrstian's thesis)

tnowotny:
The idea I remember was to measure the area under the curve for ThresholdCondition with a cutoff or weighting close to 0 (near threshold) as a proxy of how often spiking might have occurred.

tnowotny:
final thought on this today: I believe the above suggestion is approximately a scaled version of the spike count. Hence, we could just inject dthreshold/dx * (spikeN - targetSpikeN) into each neuron var x. Then, typically, the derivative is 0, 1 or -1 and we get what we used to do for the normal LIF (derivative == 1) ... which reminds me - we should probably define which side of 0 the spike condition needs to be on, i.e. it's legally (V-Theta) or (Theta-V) ... we can't allow both if we want to do this scheme.

neworderofjamie:
Just to clarify

> Hence, we could just inject dthreshold/dx * (spikeN - targetSpikeN) into each neuron var x

"dthreshold" and "dx" are the threshold and ODE for the forward system but this is injected into the adjoint variable corresponding to neuron var x?

tnowotny:
sorry, ASCII isn't great for expressing maths. We need the partial derivative d l_x / d x where l_x is a loss term depending on variables x (placeholder for any var). 
As a loss term I imagened 1/2 * (\int thresholdcondition dt - target)^2 so that the desired derivative is something like
(\int thresholdcondition dt - target) * \int (d thresholdcondition / d x) dt
- the bracket is approximately proportional to (spikeN - targetSpikeN) 
- the \int (d thresholdcondition / d x) dt would be a trial sum of d thresholdcondition/ d x 
- each x for which d thresholdcondition/ d x is non-zero gets the injection into its lambda_x dynamics equation
It's a bit speculative and slightly different to what we did before but it seems a bit more principled.
What do you think?

tnowotny:
Hum ... I kind of forgot to include the idea of a small window (by thinking it would remove itself by substituting the spike numbers); however, it doesn't. So, I think to be more sensible, one needs to include this.
Let's say I_theta = 1 if theta < epsilon
then one should use
(\int thresholdcondition*I_thresholdcondition dt - target)^2
which would carry through to the sum of d thresholdcondition/ d x 

But ... argh ... technically I_thresholdcondition depends on x so would contribute to the derivative ?!?
Maybe this needs more thought again. What a gnarly business this regularisation is.

PR #134, opened 2025-02-26
Postsynaptic model optimisation

neworderofjamie:
Minimising number of neuron variables is pretty key to performance so this implements any synapse variable whose jump is ``x + weight`` with ``inSyn``


tnowotny (approved):
I think it makes sense. I tried to think hard about whether there is anywhere else a replacement would be necessary but came up empty.

PR #137, opened 2025-03-17
Delay fixes

### LINKED ISSUE #129: Not setting max delays correctly can cause issues with loading
I trained a model with eventprop for 200 epochs. I set the max_delay_steps=62 for the synapses. I tried to load the saved weights, and while they didn't better than chance accuracy, they were significantly worse compared to what I observed at training. I realised that I didn't set max_delay_steps to 62 but to a 1000, after changing it to 62, I got the same results. Intuitively this should not be a problem, since the parameters were loaded.

neworderofjamie:
This fixes two nasty issues with delays:

* When learning delays using Eventprop, the synapse variables are ``float`` but get rounded to integer in the CUDA kernels to use as delays. However, when loading these checkpoints into an inference model, delay synapse variables are ``uint8_t`` or ``uint16_t``. Like C++, numpy truncates rather than rounds in this situation which reduces accuracy. Where arrays of delays are used, we now manually round them
* ``SynapseGroup.max_dendritic_delay_timesteps`` needs to always be 1 more than the largest actual delay value. When creating connections with ``max_delay_steps`` set manually,  ``_get_conn_max_delay`` did not take this into account

Additionally, ``EventPropCompiler`` and the ``Compiler`` base class had their own copies of ``_get_conn_max_delay`` which I have de-duplicated

Fixes #129

### DIFF
--- a/ml_genn/ml_genn/compilers/compiler.py
+++ b/ml_genn/ml_genn/compilers/compiler.py
@@ -131,18 +131,21 @@ def get_delay_type(max_delay):
     else:
         return "uint16_t"
 
-def _get_conn_max_delay(conn, delay):
-    # if maximum delay steps is specified
+def get_conn_max_delay(conn, delay):
+    # If maximum delay steps is specified
     if conn.max_delay_steps is not None:
         return 1 + conn.max_delay_steps
-    # Otherwise, if delays are specified as an array, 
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
Yes, nasty problems but hopefully all solved with this!
Looking sensible.

PR #141, opened 2025-04-09
Regularisation fixes

neworderofjamie:
Regularisation spike counts were being accumulated through validation set evaluation, resulting in large incorrect regularisation update in the first batch. This PR fixes this by zeroing forward pass spike count in reset (at beginning of batch) rather than reduction (at end of previous batch).


tnowotny (approved):
Very straightforward fix for a somewhat sneaky bug. I can confirm that with the fix ```SpikeCount``` does not accumulate in the validation phase any more and the problem of impeded learning due to inappropriate regularisation based on falsely accumulated spike counts seems fixed.

PR #157, opened 2025-06-10
Relative mean square

neworderofjamie:
This PR adds support for a mean-squared error based classification loss function ``RelativeMeanSquareError`` which tries to seperate the readout variables of incorrect classes from those of the correct class by a ``delta`` parameter.  The only smartness is in the bespoke reduction at the bottom of ``EventPropCompiler`` which simultaneously calculates the sum of first spike times (excluding neurons which don't spike) and the spike time of the true label (if it spikes).


ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -909,44 +918,83 @@ def build_neuron_model(self, pop: Population, model: NeuronModel,
...
+                        LambdaI = (A * LambdaV * (Alpha - Beta)) + (LambdaI * Beta);
+                        LambdaV *= Alpha;
+                        """
+                
+                    # On backward pass transition, update LambdaV if this is the first spike on an incorrect neuron
+                    transition_code = f"""
+                        if (fabs(backT + TFirstSpikeBack) < 1e-3*dt) {{
+                            const scalar iMinusVRecip = 1.0 / RingIMinusV[ringOffset + RingReadOffset];
+                            if(id == YTrueBack) {{
+                                LambdaV += (TFirstSpikeSumBack - (num_neurons * TFirstSpikeBack) + ((num_neurons - 1) * Delta)) * iMinusVRecip;
```
tnowotny, line 954: would it be clearer if ```TFirstSpikeTrueBack``` was used (even though it is of course the same value in this branch of the if clause)?

tnowotny (approved):
Yeah ... looks pretty tidy & I think it is all correct. The logic for the ringbuffer for IminusVRecip should be untouched/correct and the rest looks right.

PR #159, opened 2025-06-27
Added option to have k_reg per population

mbalazs98:


neworderofjamie (changes requested):
This seems very sensible. Two minor things:

1. You should update the type hint and documentation in the constructor
2. If you use a ``Sequential`` model this won't work. If ``self.reg_lambda_XXX`` is a dictionary, in the constructor you should rebuild all the keys with something like:
```python
self.reg_lambda_lower = {get_underlying_pop(p): r 
                         for p, r in self.reg_lambda_lower.items()}
````
```

PR #162, opened 2025-09-08
Individual optimisers

tnowotny:
I have re-implemented how we designate which variables are trained with which optimiser.
In essence, during ```EventPropCompiler.compile()``` we pass an optional argument ```optimisers``` through ```**kwargs``` that is a dictionary of dictionaries, e.g.
```python
optimisers={connection: {"var": optimiser,
"var2": optimiser2},
connection2: {"var": optimiser3}
}
```
This is then passed through and interpreted so that each variable is optimised by the specified optimiser. I think ```compile()``` is the right point to specify optimisers, as one could now use the same ```EventPropCompiler``` to make a network for different training setups (different optimisers and learning rates for different parts of the network).

Furthermore, I have added training with respect to neuron parameters. The optimiser is defined through the same mechanism, e.g.
```python
optimisers={population: {"var": optimiser}}
```
The code runs simple examples, training weights, delays, and neuron parameters, including just training neuron parameters. It remains to be tested whether the gradient for training neuron parameters is correct, and this does something useful in an appropriate example.

TODO:
- [x] Implement sentinel syntax for easier specifying of defaults (I think this will allow us to remove the optimiser passed to the constructor)
- [x] Unify ``Connection``/``Layer``/``Populations`` in optimiser dictionary
- [ ] Checkpointing - should be trivial
- [ ] Optimisers should be created with ``get_object`` so you can use e.g. "adam" rather than having to import ``Adam``

### DIFF
--- a/ml_genn/ml_genn/compilers/event_prop_compiler.py
+++ b/ml_genn/ml_genn/compilers/event_prop_compiler.py
@@ -1297,7 +1297,7 @@ class EventPropCompiler(Compiler):
         grad_terms = {}
         for p in learn_params:
             sym = sympy.Symbol(p)
-            o = - sum(sympy.diff(expr2, sym) + _get_lmd_sym(sym2)
+            o = - sum(sympy.diff(expr2, sym) * _get_lmd_sym(sym2)
                     for sym2, expr2 in model.dx_dt.items())
             # collect variables they might need to go into a ring buffer:
             o = _template_symbols(
--- a/ml_genn/ml_genn/compilers/event_prop_compiler.py
+++ b/ml_genn/ml_genn/compilers/event_prop_compiler.py
@@ -265,9 +265,8 @@ def _add_required_parameters(model: AutoModel, genn_model: Model, expression):
 
 
 class CompileState:
-    def __init__(self, losses, readouts, optimisers, backend_name):
-        self.losses = get_object_mapping(losses, readouts,
-                                         Loss, "Loss", default_losses)
+    def __init__(self, network: Network, losses, optimisers, 
+                 supported_matrix_type, backend_name):
         self.backend_name = backend_name
         self._neuron_reset_vars = []
         self._synapse_reset_vars = []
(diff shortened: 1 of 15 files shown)


ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -580,9 +574,21 @@ def pre_compile(self, network: Network,
         # Build list of output populations
         readouts = [p for p in network.populations
                     if p.neuron.readout is not None]
-
-        return CompileState(self.losses, readouts,
+        compile_state = CompileState(self.losses, readouts,
                             genn_model.backend_name)
+        # use explicit dictionary of optimisers if provided
+        if "optimisers" in kwargs.keys():
```
neworderofjamie, line 580: you don't need the ``keys()``
```suggestion
        if "optimisers" in kwargs:
```

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -702,18 +708,14 @@ def build_weight_update_model(self, conn: Connection,
         # Does this connection have a delay?
         has_delay = (not is_value_constant(connect_snippet.delay)
                      or connect_snippet.delay > 0)
-        
         # Does this connection have learnable delays
-        has_learnable_delay = conn in self.delay_learn_conns
+        has_learnable_delay = conn in compile_state.optimisers.keys()\
+            and "delay" in compile_state.optimisers[conn].keys()
```
neworderofjamie, line 713:
```suggestion
        has_learnable_delay = (conn in compile_state.optimisers
                               and "delay" in compile_state.optimisers[conn])
```

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -970,50 +972,86 @@ def create_compiled_network(self, genn_model, neuron_populations: dict,
...
-                cu_weight = self._create_optimiser_custom_update(
-                    f"Weight{i}", create_wu_var_ref(genn_pop, "weight"),
-                    create_wu_var_ref(genn_pop, "weightGradient"), 
-                    self._optimiser, genn_model)
+                gradient_vars = []
+                for var, o in optim.items():
+                    # If weight optimisation is required
+                    if var == "weight":
+                        # Create weight optimiser custom update
+                        # TODO: Should this be more general for any var in wum?
```
neworderofjamie, line 987: No, WUMs aren't really a concept exposed to mlGeNN users

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -1410,8 +1457,16 @@ def _build_in_hid_neuron_model(self, pop: Population,
                     "neurons defined in terms of AutoNeuronModel")
             
             # Build adjoint system from model
-            dl_dt, adjoint_jumps, saved_vars_timestep, saved_vars_spike =\
-                self._build_adjoint_system(model, False, self.reg_lambda != 0.0)
+            learn_params= [] if pop not in compile_state.optimisers \
+                else compile_state.optimisers[pop]
```
neworderofjamie, line 1461: Aren't these dictionaries?
```suggestion
            learn_params = compile_state.optimisers.get(pop, {})
```

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -1575,8 +1632,14 @@ def _build_out_neuron_model(self, pop: Population,
             example_timesteps=self.example_timesteps)
 
         # Build adjoint system from model
-        dl_dt, adjoint_jumps, saved_vars_timestep, saved_vars_spike =\
-            self._build_adjoint_system(model, True, False)
+        learn_params= [] if pop not in compile_state.optimisers \
+            else compile_state.optimisers[pop]
```
neworderofjamie, line 1636:
```suggestion
        learn_params = compile_state.optimisers.get(pop, {})
```

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -970,50 +972,86 @@ def create_compiled_network(self, genn_model, neuron_populations: dict,
...
+                    #gradient_vars.append((f"{p}Gradient", "scalar", 0.0))
+                # Create reset model for gradient variables
+                #assert len(gradient_vars) > 0
+                #zero_grad_model = create_reset_custom_update(
+                #    gradient_vars,
+                #    lambda name: create_var_ref(genn_pop, name))
+
+                # Add custom update
+                #self.add_custom_update(genn_model, zero_grad_model,
+                #                       "ZeroGradient", f"CUZeroPopGradient{i}")
```
neworderofjamie, line 1053: Yes, this is correct - the normal resetting mechanism should handle this correctly for population variables

```suggestion
```

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -970,50 +972,86 @@ def create_compiled_network(self, genn_model, neuron_populations: dict,
...
+                                       "ZeroGradient", f"CUZeroConnGradient{i}")
+                need_zero_gradient_update_group = True
+                i = i+1
+
+        # Loop through populations that require optimisers
+        i = 0
+        for c, optim in compile_state.optimisers.items():
+            if c in neuron_populations:
+                genn_pop = neuron_populations[c]
+                #gradient_vars = []
```
neworderofjamie, line 1031:
```suggestion
```

neworderofjamie (changes requested):
I'll have a go at the todo list but, in general, this looks really good

PR #171, opened 2025-11-17
Name tidy

### LINKED ISSUE #116: Better automatic naming of connections
Currently GeNN needs unique names for each connection but the current logic does not achieve this if there are multiple connections between the same pair of populations.

neworderofjamie:
Warming up with low-hanging fruit here...

Basically, naming populations/connections is helpful for debugging but previously, this sort of syntax:

```python
input = Population(SpikeInput(max_spikes=BATCH_SIZE * max_spikes),
                                  num_input)
```

meant names had to be added manually. No longer! Using https://pypi.org/project/varname/, we can extract "input" automatically in this scenario :scream:. In this PR we wrap up the correct call to this module and a counter to make sure we get unique names if e.g. we connect the same pair of populations multiple times.

Fixes #116

### DIFF
--- a/ml_genn/ml_genn/utils/unique_name.py
+++ b/ml_genn/ml_genn/utils/unique_name.py
@@ -33,7 +33,7 @@ class UniqueName:
         # Increment counter
         self._counter[name] += 1
 
-        # If this isn't the first usage, add suffic
+        # If this isn't the first usage, add suffix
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
Marvellous. I learned more Python today!
Looks good.

Issue #145, opened 2025-04-25
Try and protect against adding state to stateless classes


neworderofjamie:
Various objects in mlGeNN such as callbacks get copied around a lot meaning that, although it may **seem** sensible to store state in the class, it doesn't work. I feel there is some Python magic (perhaps meta classes) to prevent stuff being added to these classes.

PR #175, opened 2025-12-01
State management

### LINKED ISSUE #147: Optimisers should not be stateful
Currently, learning rate schedules modify the internal state of optimisers (i.e. ``Adam.alpha``). This is bad because it is not how mlGeNN generally works and it means changes modifications to e.g. learning rates persist throughout the lifetime of the ``Compiler`` object which owns a copy of the optimiser. @tnowotny improved this slightly in https://github.com/genn-team/ml_genn/commit/dbe548c42ff62688aea4d2b0644b2ce2b5824d1a which at least makes sure each ``CompiledNetwork``  the compiler creates gets its own copy **but** really optimiser state should only persist within the ``CompiledNetwork`` context manager i.e. it **should** work like:

```python
compiler = EventPropCompiler(..., optimiser=Adam(0.001))
compiled_net = compiler.compile(network)

for f in range(NUM_FOLDS):
    with compiled_net:
        # At this point compiled_net is actually loaded into GeNN and should come with fresh optimiser state
        # The GeNN model will already have the correct parameter value but this currently gets overridden by mlGeNN
        for e in range(NUM_EPOCHS):
            ...
            metrics, _  = compiled_net.train(...)
```

neworderofjamie:
One of the bits of poor design in mlGeNN 2 which has continuously been a pain point is that outwardly non-stateful metric, callback and optimiser classes are sometimes being used to store state. This _mostly_ worked ok for metrics and callbacks as callbacks have a bespoke system for storing state being recorded and the lifetime of the copies of these objects used at runtime, generally aligns to that of the ``CompiledTrainingNetwork.train``/``CompiledInferenceNetwork.evaluate`` calls. However, it would be very useful in lots of cases to have callbacks with state and optimisers have been a source of annoyance.

I was contemplating gnarly, more magical solutions but, here I have just updated callbacks, metrics and optimisers to provide a ``create_state`` method which returns a new state object which gets passed to subsequent calls. In general, this can be whatever you want it to be. Several classes use ``CompiledNetwork`` objects directly, others use ``dataclass`` objects and metrics use objects derived from an abstract base class as this seems the most sensible place to put the ``results`` property. Optimiser state is now created and destroyed with the context manager:

```
with compiled_net:
    # Do stuff
```

Fixes #147 #145

### DIFF
--- a/ml_genn/ml_genn/optimisers/optimiser.py
+++ b/ml_genn/ml_genn/optimisers/optimiser.py
@@ -22,7 +22,7 @@ class Optimiser(ABC):
 
     @abstractmethod
     def create_state(self) -> Any:
-        """Returns new optimiser-specific state object. 
+        """Returns a new optimiser-specific state object. 
         This should contain copies of any attributes of 
         the optimiser which can be modified at runtime
         """


tnowotny (approved):
It's definitely an improvement. I take it that this is also fixing the issue with progress bars always displaying the name "sparse categorical accuracy" no matter what the metric's result was?

PR #179, opened 2025-12-02
Restrict symbols available in ``global_dict`` when parsing

### LINKED ISSUE #178: Restrict symbols going into sympy.parse_expr
By default, ``sympy.parse_expr`` takes a global dict containing everything you get when you ``from sympy import *`` which means a lot of weird stuff gets imported which prevents you using lots of common names in ``AutoModel`` expressions. For neuron and synapse models, this should be culled to special functions GeNN supports and not much more.

neworderofjamie:
You can enter a lot of weird stuff into sympy expressions (https://github.com/sympy/sympy/blob/master/sympy/__init__.py gives some indication), much of which is not at all valid in the context of an ``AutoModel`` equation which meant you couldn't (for example) call variables "beta"! This PR restricts the amount of stuff that is available in these strings to what seems 'sensible'.

Fixes #178

### DIFF
--- a/ml_genn/ml_genn/utils/auto_model.py
+++ b/ml_genn/ml_genn/utils/auto_model.py
@@ -2,14 +2,26 @@ import numpy as np
 import sympy
 
 from typing import Any, MutableMapping, Optional, Tuple
-from .model import NeuronModel, SynapseModel
 from .value import Value
 
-from copy import deepcopy
-from itertools import chain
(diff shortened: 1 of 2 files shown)


tnowotny (approved):
Yeah, absolutely agree. This is much more controlled, and I can't think of any other stuff needed for Autoneuron definitions.

PR #181, opened 2025-12-03
Delay test

neworderofjamie:
Converted balazs "sequence_detectionn" model into a test...and immediately found a typo 😄 

On the bright side, we have progressed from 63% code coverage on master to a blistering 77.88% which is progress

Also, for some reason, this triggered a repeat of a previous unpleasant class of Python bug in the test suite - which was particularly fun to track down as you need to add strategic prints and run LOADS of tests.  I remember us having similar issues before but I can't find the relevant PR but, basically, _never_ do this in Python:

```python
def function(x=[]):
    x.append(12)
    # do something with x
```

as it is effectively the same as:

```python
x_default = {}
def function(x=x_default):
    x.append(12)
    # do something with x
```

Same applies to ``{}`` and, in this instance, parameter values were building up over the course of running the test suite 🤢

### DIFF
--- a/ml_genn/ml_genn/readouts/max_var.py
+++ b/ml_genn/ml_genn/readouts/max_var.py
@@ -45,17 +45,17 @@ class MaxVar(Readout):
             # Add code to update max variable and time 
             model.append_sim_code(
                 f"""
-                if ($({self.output_var_name}) > $({max_var_name})) {{
-                    $({max_var_name})= $({self.output_var_name});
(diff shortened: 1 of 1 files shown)


tnowotny (approved):
Wow, the Python trap is nasty. I would never have suspected anything like it. It highlights that Python is a pretty bad programming language for serious development, as much as it is convenient for sloppy prototyping. But the PR looks good; nice to see the tests build up.

PR #182, opened 2025-12-05
Asymmetric regularisation

tnowotny:
I have reintroduced asymmetric regularisation with reg_lambda_upper and reg_lambda_lower. I have run some example scripts, and it seems to work as advertised.
If reg_lambda is provided, it sets both lower and upper.

### DIFF
--- a/ml_genn/ml_genn/compilers/event_prop_compiler.py
+++ b/ml_genn/ml_genn/compilers/event_prop_compiler.py
@@ -633,23 +633,6 @@ class EventPropCompiler(Compiler):
                                "{\"weight\": \"adam\"} to optimise all "
                                "weights with the adam optimiser")
 
-        # Handle legacy regularisation strength definitions
-        reg_lambda_lower = genn_kwargs.pop("reg_lambda_lower", None)
-        reg_lambda_upper = genn_kwargs.pop("reg_lambda_upper", None)
-        reg_warning = False
-        if reg_lambda_lower is not None:
-            self.reg_lambda_lower = reg_lambda_lower
-            reg_warning = True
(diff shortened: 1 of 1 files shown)


ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -631,24 +632,12 @@ def __init__(self, example_timesteps: int, losses,
...
-                raise NotImplemented("Seperate 'reg_lambda_upper' and "
-                                     "'reg_lambda_lower' arguments for "
-                                     "EventPropCompiler are no longer "
-                                     "supported, please use 'reg_lambda'")
+        # If reg_lambda is specified, use it for both, reg_lambda_lower and
+        # reg_lambda_upper
+        reg_lambda = genn_kwargs.pop("reg_lambda", None)
+        if reg_lambda is not None:
+            reg_lambda_lower = reg_lambda
+            reg_lambda_upper = reg_lambda
```
neworderofjamie:
```suggestion
        # If regularisation strength is specified as a single float, use for both upper and lower
        if isinstance(reg_lambda, float):
            self.reg_lambda_lower = reg_lambda
            self.reg_lambda_upper = reg_lambda
        # Otherwise, unpack
        else:
            self.reg_lambda_lower, self.reg_lambda_upper = reg_lambda
```

neworderofjamie (changes requested):
Suggested a nicer API. The kwarg approach makes sense for deprecated stuff but makes it very hard to document the parameters. On the subject of which, could you add a description of what happens when you specify two values to the docstring on line 571?

PR #183, opened 2025-12-09
Loss recording

neworderofjamie:
After some thought and fiddling, my conclusion is that loss is a compiler-specific thing **and** needs calculating in the backward pass so recording it isn't really something that can be easily or sensibly implemented as a metric. Therefore, this PR adds functionality to generate loss-calculating terms to the event prop backward pass and stores the data in a callback added by the compiler (in the same way rewiring stats are stored with Deep-R in #94).

However,  I cannot quite figure out how all the calculations works in the original genn_eventprop code

### DIFF
--- a/ml_genn/ml_genn/compilers/event_prop_compiler.py
+++ b/ml_genn/ml_genn/compilers/event_prop_compiler.py
@@ -27,7 +27,6 @@ from ..utils.model import (CustomUpdateModel, Model, NeuronModel,
                            SynapseModel, WeightUpdateModel)
 from ..utils.snippet import ConnectivitySnippet
 
-from abc import abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field
 from itertools import chain
@@ -309,7 +308,7 @@ class CompileState:
         self.update_trial_pops = []
         self.adjoint_limit_pops_vars = []
         self.optimisers = {}
-        self.loss_recorder_callbacks = []
+        self.loss_recorder_populations = []
 
         # Build list of output populations
         readouts = [p for p in network.populations
@@ -534,15 +533,9 @@ class LossRecorderState:
     losses: list = field(default_factory=list)
 
 class LossRecorderCallback(Callback, ABC):
-    def __init__(self, pop: Population, key: str, var_name: str):
+    def __init__(self, pop: Population, key: str):
         self.pop = pop
         self.key = key
-        self.var_name = var_name
-    
-    @abstractmethod
-    def calculate_loss(self, loss, example_timesteps: int,
-                       batch_size: int):
-        pass
 
     def create_state(self, compiled_network, **kwargs):
         return LossRecorderState(compiled_network)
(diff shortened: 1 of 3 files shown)


ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -2022,8 +2115,13 @@ def _build_out_neuron_model(self, pop: Population,
                             const scalar backT = {self.example_timesteps * self.dt} - t - dt;
                             scalar drive = 0.0;
                             if (Trial > 0 && fabs(backT - {out_var_name}MaxTimeBack) < 1e-3*dt) {{
-                                const scalar g = (id == YTrueBack) ? (1.0 - Softmax) : -Softmax;
-                                drive = g / (num_batch * {self.dt * self.example_timesteps});
+                                if(id == YTrueBack) {{
+                                    {gen_record_code.substitute(n='-log(Softmax)')}
```
tnowotny, line 2119: this could also have been done at beginning of trial?

ml_genn/ml_genn/compilers/event_prop_compiler.py
tnowotny, line 2141: No loss recording for ```EndVar``` readouts?

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -2145,6 +2243,7 @@ def _build_out_neuron_model(self, pop: Population,
                             if (id == YTrueBack) {{
                                 const scalar fst = {1.01 * window_end} + TFirstSpikeBack;
                                 drive_p = (((1.0 - Softmax) / {self.softmax_temperature}) + ({self.ttfs_alpha} / (fst * fst))) / {self.batch_size};
+                                {gen_record_code.substitute(n='-log(Softmax)')}
```
tnowotny, line 2246: I *think* this should be ```-log(Softmax) - {self.ttfs_alpha} / fst``` to record the used loss with the regularisation term.

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -523,6 +528,47 @@ def on_batch_end(self, state, batch, metric_state):
...
+
+class LossRecorderCallbackSCE(LossRecorderCallbackBase):
+    def calculate_loss(self, state, loss_sum):
+        batch_size = state.compiled_network.genn_model.batch_size
+        return np.sum(loss_sum) / batch_size
+
+class LossRecorderCallbackMSE(LossRecorderCallbackBase):
+    def calculate_loss(self, state, loss_sum):
+        batch_size = state.compiled_network.genn_model.batch_size
+        return np.sum(np.sqrt(loss_sum)) / batch_size
```
tnowotny, line 570: I don't think we are actually doing RMSE, so the square root probably should not be there.
On the other hand, I think it needs a factor 1/2, so:
```python
return np.sum(loss_sum) / (2.0*batch_size)

tnowotny (commented):
I think overall it's good, but in the details there are a couple of issues as below.

ml_genn/ml_genn/compilers/event_prop_compiler.py
neworderofjamie: it could be but then this code would be spread out further and less consistent. It's only done in the timestep where max occurred so no computational difference I don't think

PR #184, opened 2026-02-04
Added a new "end_var" (EndVar) readout 

tnowotny:
Added a new "end_var" (EndVar) readout that takes the final value of the state variable as the readout.

To be used with sce_loss and using the usual Softmax procedure. This is closely modelled on how MaxVar works.


neworderofjamie (changes requested):
The semantic distinction between ``Var`` and ``EndVar`` was something I had been meaning to fix in mlGeNN 3 (I'm sure I have done it somewhere in an unfinished branch...) - nice to be beaten to it!

I've cherry-picked in my changes to mlGeNN-TF  to use ``EndVar`` where it is required

PR #188, opened 2026-02-12
Windowed readout and fixed regularisation

tnowotny:
a) removed the 1/(t+dt) term from the regularisation term (after long discussion and rigorous testing it seemed counter-productive and there are ways to construe a justification for the term without this weighting. b) Introduced an optional window for the avg_var_exp_weight readout so that the readout var is only averaged and weighted in a defined window; if the window is not specified when creating the readout, it defaults to averaging and weighting across the entire trial as previously.

If you like the windowed approach, we can also add it to avg_var. Potentially to other readouts?

### DIFF
--- a/examples/event_prop/shd_delay_shift_blend.py
+++ b/examples/event_prop/shd_delay_shift_blend.py
@@ -217,13 +217,14 @@ with compiled_net:
         labels_train = []
         blended_dataset = blend(raw_dataset, classes)
         for events, label in blended_dataset:
-            spikes_train.append(preprocess_tonic_spikes(delay(shift(events)), dataset.ordering,
-                                                    (dataset.sensor_size[0]*N_DELAY,
-                                                    dataset.sensor_size[1],dataset.sensor_size[2])))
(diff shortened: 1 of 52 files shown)


ml_genn/ml_genn/readouts/avg_var_exp_weight.py
```diff
@@ -32,10 +40,12 @@ def add_readout_logic(self, model: NeuronModel, **kwargs):
         self.output_var_type = output_var[1]
 
         # Add code to update average variable
-        scale = 1.0 / kwargs["example_timesteps"]
-        local_t_scale = 1.0 / (kwargs["dt"] * kwargs["example_timesteps"])
+        window_start = 0 if self.window_start is None else self.window_start
+        window_end = kwargs["example_timesteps"] if self.window_end is None else self.window_end
```
neworderofjamie:
```suggestion
        window_start = self.window_start or 0
        window_end = self.window_end or kwargs["example_timesteps"]
```
is the normal python idiom for this

PR #195, opened 2026-02-26
mlGeNN 3 Deep-R

neworderofjamie:
In the same not particularly exciting vein as #194, this merges deep-r support back into mlGeNN 3. However, for reasons that are not especially clear, it seems to work here supporting my view that there was a slightly mysterious typo in mlGeNN EventProp + Deep-R...

Following @tnowotny 's argument that configuration for compiling a **specific** network should be passed to ``EventPropCompiler.compile`` rather than to the ``EventPropCompiler`` constructor, I have moved the Deep-R arguments here (see latency MNIST example). Happy to hear any thoughts on nicer syntax!

The mlGeNN 2 PR also included some more flexible syntax for adding custom updates meaning rather than having to write a custom callback class to do:
```python
CustomUpdateOnBatchEndNotFirst("AbsSumReduceBatch")
```
the built in callback back classes lets you provide a 'filter' function enabling:
```python
CustomUpdateOnBatchEnd("AbsSumReduceBatch",
                       lambda batch: batch > 0)
```


tnowotny (commented):
I think I will have to largely trust you with this one. I am not sure what syntax you felt could be improved? Maybe we can have a chat irl about it.

neworderofjamie:
I think my thought was that maybe these three separate arguments could be merged some way:
```python
optimisers={"all_connections": {"weight": Adam(1e-2)}},
deep_r_conns=[hidden], 
deep_r_record_rewiring=hidden: "in_hid_rewiring"
```
although I couldn't think of _how_ when I wrote it or now :wink:

tnowotny:
Hum ... in a sense, one could argue deep_r is some sort of optimiser, so
```
optimisers={"all_connections": {"weight": Adam(1e-2)}, "hidden": "deep_r"}}
```
would be defensible.
the ```record_rewiring``` bit is more awkward; normally would want this to be a callback like any recorder. But I suspect it's needed in the compiler already ...? Which reminds me of the "record_spikes" argument for neuron populations. Maybe it belongs to the synapse pop? Though at network build it wouldn't be known whether a synapse pop would be 'deep-r-d'.

neworderofjamie:
This was shockingly easy, actually resulted in the deletion of a load of code! Now this works:
```python
compiled_net = compiler.compile(
    network, optimisers={"all_connections": {"weight": Adam(1e-2)},
                         hidden: {"connectivity": DeepR(l1_strength=1E-8,
                                                        rewiring_record_key="in_hid_rewiring")}})
```
or, with all default parameters:
```python
compiled_net = compiler.compile(
    network, optimisers={"all_connections": {"weight": "adam"}, 
                         hidden: {"connectivity": "deep_r"}})
```
additional bonus is that you can specify L1 strength per-connection

ml_genn/ml_genn/compilers/deep_r.py
```diff
@@ -0,0 +1,232 @@
...
+# **THINK** could this be written in a less "branchy" manner?
+deep_r_l1_model = {
+    "params": [("C", "scalar"), ("NumRowWords", "unsigned int")],
+    "var_refs": [("Variable", "scalar")],
+    "extra_global_param_refs": [("Inhibitory", "uint32_t*")],
+    "update_code": """
+    const uint32_t inhibitoryWord = Inhibitory[(NumRowWords * id_pre) + (id_post / 32)];
+    
+    // **NOTE** regularization is applied to GRADIENTS which are
+    // SUBTRACTED from weight by optimizer so we want to subtract
```
tnowotny: I think this should read "add"

ml_genn/ml_genn/compilers/deep_r.py
```diff
@@ -0,0 +1,232 @@
...
+    // from excitatory gradients to push weights towards zero and
+    // add to inhibitory gradients to push weights towards zero
```
tnowotny: and "subtract" for inhibitory gradients

tnowotny (approved):
I think it all looks good, and the loss recording isn't making too bulky a code. Found a detail in a comment. As you suggested, we probably need to design some tests for the loss recordings.

PR #196, opened 2026-03-04
Add polarity parameter to preprocess_tonic_spikes

### LINKED ISSUE #50: Helper function for splitting tonic datasets into spike train per polarity
Models with separate first layers for each polarity are common

mbalazs98:
Closes #50

### DIFF
--- a/ml_genn/ml_genn/utils/data.py
+++ b/ml_genn/ml_genn/utils/data.py
@@ -79,9 +79,9 @@ def preprocess_spikes(times: np.ndarray, ids: np.ndarray,
     return PreprocessedSpikes(end_spikes, times)
 
 def preprocess_tonic_spikes_separate(events: np.ndarray, ordering: Sequence[str],
-                            shape: Tuple, time_scale=1.0 / 1000.0,
-                            dt: Optional[float] = None,
-                            histogram_thresh : Optional[int] = None) -> list:
+                                     shape: Tuple, time_scale=1.0 / 1000.0,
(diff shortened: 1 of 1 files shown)


neworderofjamie:
So, two things:
1. Tonic already provides a merging transform so we don't need to duplicate this functionality
2. This function is already more than complex enough. I think it would be nicer to have a separate function that takes tonic ``events``, ``ordering`` and ``shape``, splits them by polarity and returns two seperate tonic-format event streams (as #50 suggested)

neworderofjamie (changes requested):

ml_genn/ml_genn/utils/data.py
```diff
@@ -78,6 +78,32 @@ def preprocess_spikes(times: np.ndarray, ids: np.ndarray,
     # Return end spike indices and spike times
     return PreprocessedSpikes(end_spikes, times)
 
+def preprocess_tonic_spikes_separate(events: np.ndarray, ordering: Sequence[str],
+                            shape: Tuple, time_scale=1.0 / 1000.0,
+                            dt: Optional[float] = None,
+                            histogram_thresh : Optional[int] = None) -> list:
```
neworderofjamie:
```suggestion
                                     shape: Tuple, time_scale=1.0 / 1000.0,
                                     dt: Optional[float] = None,
                                     histogram_thresh : Optional[int] = None) -> list:
```

neworderofjamie (approved):

PR #200, opened 2026-04-16
Also added windowed readout to "Var" readout. 

tnowotny:
Noticed in the process that the ```tsRingReadOffset``` was not handled correctly for per-timestep windowed readout.
Also added windowed readout to the pattern_recognition example and tweaked a bit so it actually works quite nicely.


examples/event_prop/pattern_recognition.py
```diff
@@ -7,6 +7,7 @@
 from ml_genn.compilers import EventPropCompiler
 from ml_genn.connectivity import Dense
 from ml_genn.initializers import Normal
+from ml_genn.readouts import Var
 from ml_genn.neurons import LeakyIntegrate, LeakyIntegrateFire, SpikeInput
```
neworderofjamie:
```suggestion
from ml_genn.neurons import LeakyIntegrate, LeakyIntegrateFire, SpikeInput
from ml_genn.readouts import Var
```
alphabeticise

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -1927,20 +1927,26 @@ def _build_out_neuron_model(self, pop: Population,
...
-                            drive = error / (num_batch * {self.dt * self.example_timesteps});
+                            {ro.back_windowed_readout_code(code, self.example_timesteps, self.dt)}
                         }}
                         
                         {dynamics_code}
                         """)
 
                     # Add code to fill errors into RingBuffer
+                    # **THINK** do we want to make a smaller RingBuffer and only fill within the readout
+                    # window (would also need matching indexing above)
```
neworderofjamie:
```suggestion
                    # **THINK** do we want to make a smaller RingBuffer and only fill within 
                    # the readout window (would also need matching indexing above)
```
just to reduce ugly line-wrapping in github

ml_genn/ml_genn/readouts/var.py
```diff
@@ -29,5 +29,6 @@ def get_readout(self, genn_pop, batch_size: int, shape) -> np.ndarray:
         genn_pop.vars[self.output_var_name].pull_from_device()
 
         # Return contents, reshaped as desired
+        # **THINK** do we want to not return the readout outside the window?
```
neworderofjamie, line 32: This reading is totally wrong anyway so this will be solved by my _next_ PR as Var readout will be calculated on device

ml_genn/ml_genn/compilers/event_prop_compiler.py
```diff
@@ -1910,6 +1909,7 @@ def _build_out_neuron_model(self, pop: Population,
                             scalar drive = 0.0;
                             const int tsRingOffset = (batch * num_neurons * {2 * self.example_timesteps}) + (id * {2 * self.example_timesteps});
                             if (Trial > 0) {{
+                                tsRingReadOffset--;
```
neworderofjamie: I'm not 100%  sure about this one. ``tsRingReadOffset`` is used both when output neuron dynamics require per-timestep buffering (``dyn_ts_reset_needed = True``) **and** when loss functions do. In the latter case, the ``tsRingReadOffset`` is decremented outside of any condition before the backward dynamics e.g. on line 1979. I think:

1. It doesn't matter if this gets messed with in trial 0 as the backward dynamics don't have any effect
2. It would be more consistent with the other loss functions to not have any condition on updating this

examples/event_prop/pattern_recognition.py
```diff
@@ -79,13 +80,14 @@
                                   NUM_INPUT, record_spikes=True)
     hidden = Population(LeakyIntegrateFire(v_thresh=0.61, tau_mem=TAU_MEM),
                         NUM_HIDDEN, record_spikes=True)
-    output = Population(LeakyIntegrate(tau_mem=TAU_MEM, readout="var"),
+    ro = Var(window_start=50, window_end=1000)
+    output = Population(LeakyIntegrate(tau_mem=TAU_MEM, readout=ro),
```
neworderofjamie, line 84: Obviously good that you tested it worked but I think this is maybe extra complexity this example doesn't need

examples/event_prop/pattern_recognition.py
```diff
@@ -79,13 +80,14 @@
...
                         NUM_OUTPUT)
     
     # Connections
-    Connection(input, hidden, Dense(Normal(sd=2.5 / np.sqrt(NUM_INPUT))), Exponential(TAU_SYN))
-    Connection(hidden, hidden, Dense(Normal(sd=1.5 / np.sqrt(NUM_HIDDEN))), Exponential(TAU_SYN))
-    Connection(hidden, output, Dense(Normal(sd=2.0 / np.sqrt(NUM_HIDDEN))), Exponential(TAU_SYN))
+    Connection(input, hidden, Dense(Normal(mean=0.5 / np.sqrt(NUM_INPUT), sd=1.0 / np.sqrt(NUM_INPUT))), Exponential(TAU_SYN))
+    #Connection(hidden, hidden, Dense(Normal(sd=0.5 / np.sqrt(NUM_HIDDEN))), Exponential(TAU_SYN))
+    Connection(hidden, output, Dense(Normal(sd=1.0 / np.sqrt(NUM_HIDDEN))), Exponential(TAU_SYN))
```
neworderofjamie, line 90: are these changes to improve the accuracy of this example with mlGeNN 3 or for other reasons? It seems like you got rid of the recurrence...

neworderofjamie (changes requested):

examples/event_prop/pattern_recognition.py
tnowotny: It makes sense to do this here as well as for other regression problems because the mismatch in the beginning of the trial is a potential problem, making learning worse. So I would recommend it for any regression task, and hence in the example.

examples/event_prop/pattern_recognition.py
tnowotny: Ah, yes, probably should get of the recurrency properly. All it does is destabilise learning, and it works so much better without recurrent connections. If we want an example where actual temporal processing with recurrent connection is truly going on we need to start from scratch.

ml_genn/ml_genn/compilers/event_prop_compiler.py
tnowotny: I am happy with updating at all times. The change I made here was to avoid it being updated only within the window, which would have been properly wrong.

ml_genn/ml_genn/readouts/var.py
tnowotny: ok ... I'll let you get on with that.

ml_genn/ml_genn/compilers/event_prop_compiler.py
neworderofjamie: Ahh yes, updating only within the window definitely is bad. I've moved them out of the ``if(Trial > 0)`` for consistency

examples/event_prop/pattern_recognition.py
neworderofjamie: Performance is still fairly awful compared to the original e-prop version...
image

neworderofjamie:
Ok, this was made off an out-of-date branch so I have:

1. Rebased
2. Implemented ``EndVar`` recording (based on comments from closed PR #183)
3. Corrected python-side recording maths for MSE as they are indeed not RMSE (based on comments from closed PR #183)
4. Corrected value being recorded with sparse categorical crossentropy on spike times with bias (based on comments from closed PR #183)

neworderofjamie (approved):

examples/event_prop/pattern_recognition.py
tnowotny: the firing rates look very high; with higher hidden firing rates the performance can be further improved if that is what we want

PR #203, opened 2026-05-29
EPropCompiler: add error_start_timestep and weight_clamp parameters

Bekkozhan:
Implements two features discussed in #753:

1) error_start_timestep — restricts the error signal to a specific time window (e.g. recall window in the evidence accumulation task). As suggested by @neworderofjamie.
2) weight_clamp — exposes the existing optimiser clamping infrastructure to users via a dict of connection → (min, max) bounds. Validated on the evidence accumulation task: mlGeNN reaches 91.1% final accuracy (comparable to NEST's 93.8%) with stable convergence over 900 epochs.

### DIFF
--- a/ml_genn/ml_genn/compilers/eprop_compiler.py
+++ b/ml_genn/ml_genn/compilers/eprop_compiler.py
@@ -278,6 +278,11 @@ class EPropCompiler(Compiler):
         f_target:                   Target hidden neuron firing rate used for
                                     regularisation [Hz]
         train_output_bias:          Should output neuron biases be trained?
+        error_start_timestep:       Timestep from which the error signal is applied
+                                    to output neurons. Default 0 applies the error
+                                    from the start of the trial. Set to e.g. 2100
+                                    for the evidence accumulation task to restrict
+                                    learning to the recall window only [timesteps]
         dt:                         Simulation timestep [ms]
         batch_size:                 What batch size should be used for
                                     training? In our experience, e-prop works
@@ -302,6 +307,7 @@ class EPropCompiler(Compiler):
     def __init__(self, example_timesteps: int, losses, optimiser="adam",
                  tau_reg: float = 500.0, c_reg: float = 0.001, 
                  f_target: float = 10.0, train_output_bias: bool = True,
+                 error_start_timestep: int = 0,
                  dt: float = 1.0, batch_size: int = 1,
                  rng_seed: int = 0, kernel_profiling: bool = False,
                  reset_time_between_batches: bool = True,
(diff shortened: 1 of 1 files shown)


ml_genn/ml_genn/compilers/eprop_compiler.py
```diff
@@ -387,7 +396,12 @@ def build_neuron_model(self, pop: Population, model: NeuronModel,
             # Add sim-code to calculate error
             model_copy.append_sim_code(
                 f"""
-                E = {model_copy.output_var_name} - yTrue;
+                if(t >= {self.error_start_timestep * self.dt}) {{
+                    E = {model_copy.output_var_name} - yTrue;
+                }}
+                else {{
+                    E = 0.0;
+                }}
```
neworderofjamie: would be nice to only generate this code if you only generated this code if ``self.error_start_timestep != 0``

ml_genn/ml_genn/compilers/eprop_compiler.py
```diff
@@ -635,7 +649,9 @@ def create_compiled_network(self, genn_model, neuron_populations: dict,
             compile_state.checkpoint_population_vars, self.reset_time_between_batches)
 
     def _create_optimiser_custom_update(self, name_suffix, var_ref,
-                                        gradient_ref, genn_model, wu):
+                                        gradient_ref, genn_model, wu, conn=None):
```
neworderofjamie: I think it would be nicer if ``clamp`` was determined at the call-site and passed through to this function rather than ``conn``

ml_genn/ml_genn/compilers/eprop_compiler.py
```diff
@@ -325,12 +332,14 @@ def __init__(self, example_timesteps: int, losses, optimiser="adam",
...
         self.f_target = f_target
         self.train_output_bias = train_output_bias
+        self.error_start_timestep = error_start_timestep
         self.reset_time_between_batches = reset_time_between_batches
         self.deep_r_conns = set(get_underlying_conn(c) for c in deep_r_conns)
         self.deep_r_l1_strength = deep_r_l1_strength
         self.deep_r_record_rewirings = {get_underlying_conn(c): k
                                         for c, k in deep_r_record_rewirings.items()}
-
+        self._weight_clamp = {get_underlying_conn(c): bounds
```
neworderofjamie: no good reason but, just for consistency, it would be good to remove the underscore and just call it ``weight_clamp``

neworderofjamie (changes requested):
Looks great, just a few very minor suggestions

neworderofjamie (approved):

PR #204, opened 2026-06-15
Add 'predict' method to FewSpikeCompiler

danilo-bc:
I modified FewSpikeCompiler in `ml_genn/ml_genn/compilers/few_spike_compiler.py` and added `predict` and ` predict_batch_iter` as copies from `evaluate` and `evaluate_batch_iter`.

The main addition is a simple Python list which aggregates the last `batch_y_pred` to be returned as numpy array.

The `examples/tf/simple_cnn.py` example was used to test the functionality, so I appended a couple lines of manual accuracy calculation to it. Here are the last lines of the console after execution.

```bash
313/313 [==============================] - 2s 4ms/step - loss: 0.0307 - accuracy: 0.9906
TF evaluation time: 1.824444
make: Entering directory '/<root>/ml_genn/examples/tf/simple_cnnFewSpikeCompiler_CODE'
make: Leaving directory '/<root>/ml_genn/examples/tf/simple_cnnFewSpikeCompiler_CODE'
100%|████████▉| 1116/1117 [00:02<00:00, 540.30it/s, SparseCategoricalAccuracy: 0.9906]
Accuracy from metrics = 99.06%
FewSpikeCompiler 'evaluate' time = 2.086701174001064 s
100%|████████▉| 1116/1117 [00:02<00:00, 542.13it/s, SparseCategoricalAccuracy: 0.9906]
Accuracy from manual calc. from predict = 99.06%
FewSpikeCompiler 'predict' time = 2.063075926998863 s
```

### Minor issues or non-issues/outside of scope that were not changed

- My first idea was to copy the structure from `InferenceCompiler`, which seems to be a more mature version of what was available in `FewSpikeCompiler`. The accuracy calculation in `evaluate` came out with bad results (from ~99% to ~9%) and I couldn't quickly trace and solve the problem. Maybe it has to do with the `deque`s in FewSpikeCompiler, which appropriately sync outputs and steps.
- I noticed an unused `evaluate_batch_iter` in `ml_genn/ml_genn/compilers/inference_compiler.py`, but I don't want to be removing things outside of the scope of the feature by my own.
- The counter for the progress bar in `evaluate`/`predict` goes beyond the total number of expected batches, because there are some batches required to sync the pipeline with the outputs. I believe that's fine, since those extra batches *are* run during the prediction process.
- I didn't neatly include `predict` as an internal step of `evaluate` because a tidier structure was already done in InferenceCompiler, but I couldn't import from that class.
- I didn't remove the metrics calculation and update in `predict`, because  it also participates in the callback update loop (`callback_list.on_batch_end(batch_i, metrics)`) but it looks like it can be removed.

Please feel free to edit or suggest editions to the code in this PR, I'm happy to contribute.


ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
         # Return metrics
         return metrics, callback_list.get_data()
 
+    def predict(self, x: dict, y: dict,
+                 metrics="sparse_categorical_accuracy",
```
neworderofjamie:
```suggestion
    def predict(self, x: dict, outputs: Union[Sequence, PopulationType],
```
Predict should have the same signature as it does in ``InferenceCompiler`` - labels and metrics not required

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+            x:          Dictionary of inputs to inject 
+                        into input neuron populations.
+            y:          Dictionary of labels to compare to
+                        readout from output neuron population.
+            metrics:    Metrics to calculate.
+            callbacks:  List of callbacks to run during evaluation.
+        """
+        # Determine the number of elements in x and y
+        x_size = get_dataset_size(x)
+        y_size = get_dataset_size(y)
```
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+
+        if x_size is None:
+            raise RuntimeError("Each input population must be "
+                               " provided with same number of inputs")
+        if y_size is None:
+            raise RuntimeError("Each output population must be "
+                               " provided with same number of labels")
+        if x_size != y_size:
+            raise RuntimeError("Number of inputs and labels must match")
```
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+
+        # Batch x and y
+        # [[in_0_batch_0, in_0_batch_1], [in_1_batch_1, in_1_batch_1]]
+        splits = range(0, x_size, self.genn_model.batch_size)
+        x_batched = [[d[s:s + self.genn_model.batch_size] for s in splits]
+                     for d in x.values()]
+        y_batched = [[d[s:s + self.genn_model.batch_size] for s in splits] 
+                     for d in y.values()]
```
neworderofjamie:
```suggestion
        # Convert outputs to sequence
        outputs = outputs if isinstance(outputs, Sequence) else [outputs]
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+
+        # Zip together and evaluate using iterator
+        return self.predict_batch_iter(list(x.keys()), list(y.keys()),
+                                        iter(zip(*(x_batched + y_batched))),
+                                        len(splits), metrics, callbacks)
+
+    def predict_batch_iter(self, inputs, outputs, data: Iterator,
+                            num_batches: Optional[int] = None,
+                            metrics="sparse_categorical_accuracy",
```
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+            metrics:    Metrics to calculate.
+            callbacks:  List of callbacks to run during evaluation.
+        """
+        # Convert inputs and outputs to tuples
+        inputs = inputs if isinstance(inputs, Sequence) else (inputs,)
+        outputs = outputs if isinstance(outputs, Sequence) else (outputs,)
+
+        # Build metrics
+        metrics = get_object_mapping(metrics, outputs, Metric, 
+                                     "Metric", default_metrics)
```
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+                    batch_y_true = y_pipe_queue[o].popleft()
+
+                    # Get predictions from model
+                    batch_y_pred = self.get_readout(o)
+                    for y_pred_item in batch_y_pred[:len(batch_y_true)]:
+                        all_y_pred.append(np.copy(y_pred_item))
+                    # Update metrics
+                    metrics[o].update(batch_y_true,
+                                      batch_y_pred[:len(batch_y_true)],
+                                      self.communicator)
```
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+
+            # End batch
+            callback_list.on_batch_end(batch_i, metrics)
```
neworderofjamie:
```suggestion
            callback_list.on_batch_end(batch_i, {})
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,6 +184,157 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+
+            # Next batch
+            batch_i += 1
+
+        # End testing
+        callback_list.on_test_end(metrics)
```
neworderofjamie:
```suggestion
        callback_list.on_test_end({})
```

neworderofjamie (changes requested):
Thank you so much for this! Some tidying may be required but this is my attempt at removing the unnecessary bits

danilo-bc:
I applied the suggestions, but some things broke because of the change in interface. This is specially because of the call to `self.predict_batch_iter` and the new interface. Something about changing the data iterator to `iter(x_batched)` instead of the zip of x + y didn't go well. I'll review this and some other things to get it to a working state.

neworderofjamie:
yeah, I couldn't quite figure out what you'd done there so I left it alone - probably just copy what happens in ``InferenceCompiler``

danilo-bc:
It's funny that my first attempt was to just copy what happened in `InferenceCompiler` but "evaluate" didn't run properly. Then, this attempt is copying `evaluate` adding an output and removing the metrics things. Maybe I passed the wrong arguments in the main script for the interface from `InferenceCompiler`. I'll review once again later.

danilo-bc:
I tried once again copying all `evaluate` and `predict` methods from InferenceCompiler, but they don't work with the example as it is. Evaluate returns 9% accuracy in a 99% base model with the code taken directly from InferenceCompiler.

The [pipeline part of the current FewSpikeCompiler code](https://github.com/danilo-bc/ml_genn/blob/fda68965fde35f115a6f9c0d82e518465b5e66cf/ml_genn/ml_genn/compilers/few_spike_compiler.py#L280) seems to be important to synchronize read out outputs with the pipeline depth.

The current iteration has the same interface as InferenceCompiler, also returning a dictionary containing the mapping of outputs and their predicted data. Docstrings were left open until everything settles.

neworderofjamie:
The pipeline part is indeed very important to synchronise inputs and outputs. Just to clarify, does the current version predict correctly or not?

danilo-bc:
The current version matches the results from the untouched `evaluate`. This is the terminal output of running `python simple_cnn.py --batch-size 9 --reuse-tf-model`

```bash
313/313 [==============================] - 2s 4ms/step - loss: 0.0307 - accuracy: 0.9906
TF evaluation time: 1.755644
make: Entering directory '/<root>/ml_genn/examples/tf/simple_cnnFewSpikeCompiler_CODE'
make: Leaving directory '/<root>/ml_genn/examples/tf/simple_cnnFewSpikeCompiler_CODE'
100%|██████████▉| 1116/1117 [00:01<00:00, 586.56it/s, SparseCategoricalAccuracy: 0.9906]
Accuracy from metrics = 99.06%
FewSpikeCompiler 'evaluate' time = 1.927049159001399 s
100%|██████████▉| 1116/1117 [00:01<00:00, 617.57it/s]
Accuracy from manual calc. from predict = 99.06%
FewSpikeCompiler 'predict' time = 1.8091718630003015 s
```

The pipeline in `evaluate` is important, but the pipeline in `predict` can be changed into a simple counter, I think. There's some cleanup to be done on docstrings.

danilo-bc:
Hi. I reviewed the changes using `examples/tf/simple_cnn.py` with the addition of `predict` to that example and everything is still matching.

```python
100%|██████████▉| 1116/1117 [00:02<00:00, 479.56it/s, SparseCategoricalAccuracy: 0.9904]
Accuracy from metrics = 99.03999999999999%
FewSpikeCompiler 'evaluate' time = 2.349788557003194 s
100%|█████████████████████████████████████████████▉| 1116/1117 [00:02<00:00, 455.96it/s]
Accuracy from manual calc. from predict = 99.03999999999999%
FewSpikeCompiler 'predict' time = 2.4502199170019594 s
```

Please let me know if we need additional changes

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -66,7 +65,7 @@ def evaluate(self, x: dict, y: dict,
         # Batch x and y
         # [[in_0_batch_0, in_0_batch_1], [in_1_batch_1, in_1_batch_1]]
         splits = range(0, x_size, self.genn_model.batch_size)
-        x_batched = [[d[s:s + self.genn_model.batch_size] for s in splits]
+        x_batched = [[d[s:s + self.genn_model.batch_size] for s in splits]
```
neworderofjamie:
```suggestion
        x_batched = [[d[s:s + self.genn_model.batch_size] for s in splits]
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -76,18 +75,19 @@ def evaluate(self, x: dict, y: dict,
...
 
-    def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
-                            num_batches: Optional[int] = None,
-                            metrics="sparse_categorical_accuracy",
-                            callbacks=[BatchProgressBar()]):
-        """ Evaluate an input in iterator format against labels
+    def evaluate_batch_iter(
+            self, inputs, outputs, data: Iterator, num_batches: int = None,
+            metrics: MetricsType = "sparse_categorical_accuracy",
+            callbacks=[BatchProgressBar()]):
```
neworderofjamie: Can you revert these formatting changes? The correct type hint for an optional integer is ``Optional[int]`` not ``int`` and  the previous formatting was preferable

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -40,15 +40,14 @@ def __init__(self, genn_model, neuron_populations,
...
-            x:          Dictionary of inputs to inject 
-                        into input neuron populations.
-            y:          Dictionary of labels to compare to
-                        readout from output neuron population.
+            x:          Dictionary of testing inputs
+            y:          Dictionary of testing labels to compare 
+                        predictions against
             metrics:    Metrics to calculate.
-            callbacks:  List of callbacks to run during evaluation.
+            callbacks:  List of callbacks to run during inference.
```
neworderofjamie: can you revert these docstring changes? The previous docstrings weren't amazing but these don't improve anything and are unrelated to this PR

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +184,121 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+        x_batched = batch_dataset(x, self.genn_model.batch_size, x_size)
+        
+        # Zip together and evaluate using iterator
+        return self.predict_batch_iter(x_batched, outputs, x_size,
+                                        len(splits), callbacks)
+
+    def predict_batch_iter(self, inputs, outputs, x_size,
+                            num_batches: Optional[int] = None,
+                            callbacks=[BatchProgressBar()]):
+        """ Predict an input in iterator format against labels
```
neworderofjamie: The signature here isn't quite right. The idea is that this can take an iterator to e.g. a Tensorflow dataset. Like ``evaluate_batch_iter``, it should take ``inputs`` (which is a single input population/sequence of input populations),  ``outputs`` (which is a single output population/sequence of output populations) and ``data`` which is a data iterator which returns batches of inputs for each input.

examples/tf/simple_cnn.py
```diff
@@ -103,8 +103,17 @@
...
+    print(f"FewSpikeCompiler 'evaluate' time = {end_time - start_time} s")
+
+    start_time = perf_counter()
+    y_pred_dict, cb_data = compiled_net.predict(
+        {net_inputs[0]: validate_x}, net_outputs,
+        callbacks=callbacks)
+    end_time = perf_counter()
+    y_pred_accuracy = np.count_nonzero(np.argmax(y_pred_dict[net_outputs[0]], axis=1) == validate_y) / len(validate_y)
+    print(f"Accuracy from manual calc. from predict = {100.0 * y_pred_accuracy}%")
+    print(f"FewSpikeCompiler 'predict' time = {end_time - start_time} s")
```
neworderofjamie: please revert these changes.

neworderofjamie (changes requested):
Thanks for tidying up the prediction implementation. Only outstanding issue is that ``predict_batch_iter`` has the wrong signature to work with iterable datasets. Simplest way to test is using https://github.com/genn-team/ml_genn/blob/master/examples/tf/vgg16_cifar10.py and pre-trained weights from https://doi.org/10.25377/sussex.17380853.

It has been a while since I looked at this but:

```python
def cifar10_dataset_predict():
    _, (validate_x, _) = tf.keras.datasets.cifar10.load_data()
    validate_x = validate_x.astype('float32')

    color_normalize_fn = color_normalize(cifar10_mean, cifar10_std)

    validate_ds = tf.data.Dataset.from_tensor_slices((validate_x,))
    validate_ds = validate_ds.map(color_normalize_fn, num_parallel_calls=tf.data.AUTOTUNE)

    return validate_ds
```

should give you a suitable iterator to use.

ml_genn/ml_genn/compilers/few_spike_compiler.py
danilo-bc: I extracted that from inference_compiler.py, for consistency
https://github.com/genn-team/ml_genn/blob/4cd6cb6728d016687c7216f11de7d20250fba21c/ml_genn/ml_genn/compilers/inference_compiler.py#L146 

as I tried to copy as much of the inference interface as I could. I also see that `metrics` has `MetricsType` in the new version, but simply `"sparse_categorical_accuracy"` in the previous one.

Should I review both files to use `Optional[int]` and `MetricsType = "sparse_categorical_accuracy"`?

ml_genn/ml_genn/compilers/few_spike_compiler.py
danilo-bc: I tried to follow inference compiler and make them a single docstring, since they should work the same.

https://github.com/genn-team/ml_genn/blob/4cd6cb6728d016687c7216f11de7d20250fba21c/ml_genn/ml_genn/compilers/inference_compiler.py#L40

Should I still revert?

ml_genn/ml_genn/compilers/few_spike_compiler.py
danilo-bc: I'll take my time to review this to get the interface properly. Here are some observations from my train of thought to try to get everything right.

Taking inspiration from inference_compiler.py, again, I really tried to simply import evaluate/predict methods from it, since it seems to attend to the main requirements of this implementation (interfaces, methods etc).

The chain of processing for `CompiledInferenceNetwork` when `predict` is called starts with this method:
```python
def predict(self, x: dict, outputs: Union[Sequence, PopulationType],
                callbacks=[BatchProgressBar()]):
```
Which calls
```python
# Loop through batches and evaluate
for batch, x_batch in enumerate(x):
    # [irrelevant callback code]
    # Get predictions from each output on this batch
    y_pred_batch = self._predict_batch(batch, x_batch, outputs,
                                   callback_list)
```
Whose signature is
```python
def _predict_batch(self, batch: int, x: dict, outputs: Sequence,
                       callback_list: CallbackList):
```
I thought this was a clean implementation which separates batch loop control and singular batch prediction. The main **pitfall** of this solution seems to be the **lack of either a queue or a pipeline counter to synchronize the appropriate outputs**.

The current solution in `few_spike_compiler.py` uses a few of the utilities in `inference_compiler.py`, but without the neat loop element separation in a method.

Isolating only the flow of input data, we have
```python
# predict
x_batched = batch_dataset(x, self.genn_model.batch_size, x_size) # L 205
return self.predict_batch_iter(x_batched, outputs, x_size,       # L 208
                                        len(splits), callbacks)
# 'x_batched' becomes 'inputs'
# predict_batch_iter
inputs = inputs if isinstance(inputs, Sequence) else (inputs,) # L 222
data = iter(inputs)                                            # L 246
batch_input_x = next(data)                                     # L 254
self.set_input(batch_input_x)                                  # L 261

## After synchronizing pipeline
y_pred_batch = self.get_readout(o)                             # L 283
```

From checking `vgg16_cifar10.py`, I see evaluating from an iterator is important (as you mentioned yourself). I'll refactor the creation of the iterator (`data = iter(inputs)`) to the `predict` function itself so I can get the signature of `predict_batch_iter` right and use that Python script as validation that it works as intended.

Please let me know if this makes sense and if there's any script with a different usage scenario to validate.

examples/tf/simple_cnn.py
danilo-bc: Should I make any other script in the `examples` folder showcasing the `predict` functionality? It would be the first example of this for `ml_genn_tf`.

ml_genn/ml_genn/compilers/few_spike_compiler.py
neworderofjamie: Like I say, this change is unrelated to the purpose of this PR so please just revert it. IMO, both the formatting and the type hint in the few-spike compiler were better but it's not super-important

ml_genn/ml_genn/compilers/few_spike_compiler.py
neworderofjamie: again, clearly there are inconsistencies, but this isn't related to the purpose of this PR

examples/tf/simple_cnn.py
neworderofjamie: I don't think it's necessary

ml_genn/ml_genn/compilers/few_spike_compiler.py
neworderofjamie: the _algorithm_ with the counter you have used in your new ``predict_batch_iter`` implementation is correct. The signature just needs to be:

```python
def predict_batch_iter(self, inputs, outputs, data: Iterator,
                       num_batches: Optional[int] = None,
                       callbacks=[BatchProgressBar()]):
```

so it can be used with iterator datasets such as the one used in vgg16_cifar.py

danilo-bc:
All points were taken into consideration. I could successfully run vgg16_cifar10.py using both `evaluate_batch_iter` and `predict_batch_iter`, and then use `argmax` in the predictions output to confirm the results are matching!

I ended up separating `predict` from `predict_batch_iter` because the interfaces became far too different to use one inside the other, while making things unnecessarily complicated, even with some repeated code.

The extra docstring modifications in `evaluate_batch_iter` are because the original docstring was taken directly from `evaluate`, even with a different interface.

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +186,206 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+            
+            callback_list.on_batch_end(batch_i, {})
+            batch_i += 1        
+        callback_list.on_test_end({})
+        # Concatenate predictions into single numpy array and trim padding
+        for o in outputs:
+            y_pred[o] = np.concatenate(y_pred[o])[:x_size,:]
+        
+        # Return predictions and metrics
+        return y_pred, callback_list.get_data()
```
neworderofjamie: I don't understand why duplicating this logic is necessary. Can't you just do this?
```suggestion
        # Zip together and evaluate using iterator
        return self.predict_batch_iter(list(x.keys()), outputs,
                                        iter(zip(*x_batched),
                                        len(splits), callbacks)
```

neworderofjamie (changes requested):

ml_genn/ml_genn/compilers/few_spike_compiler.py
danilo-bc: If we really want one to use the other, I'll try to make it work. When I tried to do both at the same time, some lines weren't as straightforward, like the `set_input`, `x_size` definition, and the `next(data)` parts.

It's certainly doable.

neworderofjamie: I don't really understand what you mean. I think you literally need to make the change I suggested - this is exactly how ``evaluate``/``evaluate_batch_iter`` work

danilo-bc: The latest version removes the duplication, but I add a fake `y_batched` to make the `predict` and TF iterators match, so I can use `next` and pick the input data easily.

neworderofjamie: If you use the ``cifar10_dataset_predict`` function I pasted above to generate a TF iterator, is this necessary?

danilo-bc: OK, I understand where this is going and I just tested the function above. It fails on `.map`, but I think I can fix it and take the iterator with just `validate_x`. If all goes right, I'll correct the use of `next` to only expect new `input_batch`, and no need to exclude the labels using `_`.

danilo-bc: I'm not sure if I'm missing something simple or if Python is showing different behavior. Now **vgg16_cifar10.py** works, but **simple_cnn.py** doesn't.

I tried your original `cifar10_dataset_predict()`. It throws an error because `color_normalize_fn` expects `label` as well. I did a modified version to properly obtain only validate_x information in the iterator:

```python
def cifar10_dataset_predict():
    _, (validate_x, validate_y) = tf.keras.datasets.cifar10.load_data()
    validate_x = validate_x.astype("float32")

    color_normalize_fn = color_normalize(cifar10_mean, cifar10_std)

    ds = tf.data.Dataset.from_tensor_slices((validate_x, validate_y))
    ds = ds.map(color_normalize_fn, num_parallel_calls=tf.data.AUTOTUNE)

    # Remove the labels after normalization
    validate_ds = ds.map(lambda x, y: x, num_parallel_calls=tf.data.AUTOTUNE)

    return validate_ds
```

Then, an iterator is created using the same procedure

```python
     validate_predict_ds = cifar10_dataset_predict()

    if args.n_test_samples is None:
        args.n_test_samples = 10000

    mlg_validate_predict_ds = validate_predict_ds.take(args.n_test_samples)
    mlg_validate_predict_ds = mlg_validate_predict_ds.batch(args.batch_size)
    mlg_validate_predict_ds = mlg_validate_predict_ds.prefetch(tf.data.AUTOTUNE)
```

and a prediction is called as
```python
        y_pred_dict, cb_data = compiled_net.predict_batch_iter(
            net_inputs, net_outputs, iter(mlg_validate_predict_ds),
            num_batches=num_batches, callbacks=callbacks)
```

I believe this is the desired behavior for the TF case.

Currently, the `next(data)` applied to the TF example returns `batch_size` samples, whereas coming from `predict` returns a batch with size 1, since we grouped them by input population first.

This is why I thought this line looked weird.
```
self.set_input({inputs[0]: input_batch})
```
For `predict`, we really want the association of `input[i]` and `input_batch[i]`, but for TF, we currently have no distinction of input populations and input batches.

I may not have the full context, but I'm thinking the solution path is to group TF iterators per input, then change `predict_batch_iter` to take it into consideration.

danilo-bc:
I think all main points were achieved with this last commit. I closed the previous thread because the code is in a new state for review. Please let me know if I need to change anything else.

As is, the accuracy between `evaluate` and `predict` for both `simple_cnn.py` and `vgg16_cifar10.py` examples are the same.

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +186,142 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+        
+        # Create callback list and begin testing
+        num_batches = (None if num_batches is None
+                       else num_batches + 1 + max(y_pipe_depth.values()))
+        callback_list = CallbackList(callbacks, compiled_network=self,
+                                     num_batches=num_batches)
+        callback_list.on_test_begin()
+        
+        # Counter to synchronize outputs with pipeline
+        y_pipe_counter = {p: 0 for p, _ in y_pipe_depth.items()}
```
neworderofjamie:
```suggestion
        y_pipe_counter = {p: 0 for p in y_pipe_depth.keys()}
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +186,142 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+                    # full_input_size is used to crop output
+                    # It won't work if inputs from different populations
+                    # have different sizes
+                    full_input_size += len(input_batch) 
+                # Counter to match output readout with network pipeline latency
+                if len(outputs) == 1:
+                    y_pipe_counter[outputs[0]] += 1
+                else:
+                    for p in outputs:
+                        y_pipe_counter[p] += 1
```
neworderofjamie: I am fairly certain the special case isn't required here
```suggestion
                for p in outputs:
                    y_pipe_counter[p] += 1
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +186,142 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+            
+            if data_remaining:
+                # Set input from batch
+                # If input_batch is already organized as dict, set as is
+                if isinstance(input_batch, dict):
+                    self.set_input(input_batch)
+                    # full_input_size is used to crop output
+                    # It won't work if inputs from different populations
+                    # have different sizes
+                    full_input_size += len(input_batch[inputs[0]])
```
neworderofjamie: what is this case for?

ml_genn/ml_genn/compilers/few_spike_compiler.py
neworderofjamie: As I say above, I don't think  the ``full_input_size`` logic really works (as you say because it doesn't handle different sized inputs)
```suggestion
                # Add number of elements in output batch to correct queue(s)
                if len(outputs) == 1:
                    y_pipe_queue[outputs[0]].append(len(input_batch))
                else:
                    for p, x in zip(outputs, input_batch):
                        y_pipe_queue[p].append(len(x))
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +186,142 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+        callback_list = CallbackList(callbacks, compiled_network=self,
+                                     num_batches=num_batches)
+        callback_list.on_test_begin()
+        
+        # Counter to synchronize outputs with pipeline
+        y_pipe_counter = {p: 0 for p, _ in y_pipe_depth.items()}
+        
+        data_remaining = True
+        batch_i = 0
+        full_input_size = 0
```
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
neworderofjamie:
```suggestion
```

ml_genn/ml_genn/compilers/few_spike_compiler.py
```diff
@@ -184,13 +186,142 @@ def evaluate_batch_iter(self, inputs, outputs, data: Iterator,
...
+                    # **YUCK** this isn't quite right as input_batch
+                    # could also have outer dimension
+                    if len(inputs) == 1:
+                        self.set_input({inputs[0]: input_batch})
+                    else:
+                        self.set_input({p: x for p, x in zip(inputs, input_batch)})
+                    # full_input_size is used to crop output
+                    # It won't work if inputs from different populations
+                    # have different sizes
+                    full_input_size += len(input_batch)
```
neworderofjamie:
```suggestion
```

danilo-bc:
Bumping this PR thread to understand if we need any more changes. Please let me know if something is off with this final iteration.

PR #206, opened 2026-09-04
prefer_in_memory

neworderofjamie:
* replaced previous weird ``prefer_in_memory`` flag to inference and few-spike compilers with ``optimise_connectivity_speed`` which better reflects its role and also has more sensibly ordered priority lists
* updated tests
* updated TF examples

``optimise_connectivity_speed`` remains a bit of a mouthful so would welcome terser suggestions!


tnowotny (approved):
That looks more understandable and I agree with the default False, so that mempry savings are prioritised by default.

